PR #102, opened 2023-01-20
Add term attributes / Avoid calling the LLVM simplification

### LINKED ISSUE #88: Synthetic variable support
If we are going to call the LLVM code to simplify the term multiple times, we might need to cache which parts of the term are concrete. This will probably require a new representation for terms, ideas:
- Cofree (like in the old backend)
- a plain old product type (basically a custom Cofree)
- GADTs to mark concrete terms (every term could have a tag? you only look at the constructor to see if a Term is symbolic or concrete)

Something else to think about:
```
syntax Int ::= f(Int) [no-evaluators]

f(3)  ----LLVM----> f(3)

old backend ----> constructor-like, f(3) is not constructor-like
```

### LINKED ISSUE #101: Simplify terms on-demand instead of at every rewrite step
The pattern's term should not be simplified at every step. Instead, 
1. simplify before starting to rewrite
2. simplify and re-try when a rewrite step returns indeterminate because of "unification unclear"
   - could be more specific with a new failure type
   - followed by retrying (once!) the rewrite step
   - second failure will abort the rewrite as before
3. simplify the returned pattern's term when the rewrite ends (in any case)

jberthold:
We only call the LLVM simplification when rewriting cannot proceed (stuck or unification unclear), and before we return the result to the caller.
This speeds up the `IMP` execution quite a bit (together with an optimisation using the attributes in the substitution routine).

Includes runtimeverification/hs-backend-booster#100 
Fixes runtimeverification/hs-backend-booster#88 
Fixes runtimeverification/hs-backend-booster#101

### DIFF
--- a/library/Kore/JsonRpc.hs
+++ b/library/Kore/JsonRpc.hs
@@ -80,7 +80,7 @@ respond def@KoreDefinition{} mLlvmLibrary =
 
         _ -> pure $ Left $ ErrorObj "Not implemented" (-32601) Null
   where
-    execResponse :: (Natural, RewriteResult) -> Either ErrorObj (API 'Res)
+    execResponse :: (Natural, RewriteResult Pattern) -> Either ErrorObj (API 'Res)
     execResponse (_, RewriteSingle{}) =
         error "Single rewrite result"
     execResponse (d, RewriteBranch p nexts) =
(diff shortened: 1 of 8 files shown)


goodlyrottenapple (approved):
LGTM

PR #191, opened 2023-06-14
rename booster to kore-rpc-booster

goodlyrottenapple:


test/rpc-integration/default.nix
```diff
@@ -15,7 +15,7 @@ let
         ${lib.strings.concatMapStrings (f: ''
           export ${f}
         '') buildFlags}
-        export SERVER=${booster}/bin/booster
+        export SERVER=${kore-rpc-booster}/bin/kore-rpc-booster
```
jberthold, line 18: We also need to adjust the script in `test/rpc-integration` , these variables are mainly for nix.

jberthold (approved):

PR #195, opened 2023-06-16
194 add length field to binary kore format

jberthold:
Kore [version 1.2.0 introduces an 8-byte length field](https://github.com/runtimeverification/llvm-backend/commit/90ff252b58fe7727eed0c0afdad3b30d37766f62) (as part of the header).
* check that the version is supported (range 1.0 .. 1.2)
* read the length field if version >= 1.2
* calculate total byte length and check it while decoding the data block

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -48,51 +48,6 @@
         "type": "github"
       }
     },
-    "blank": {
-      "locked": {
-        "lastModified": 1625557891,
-        "narHash": "sha256-O8/MWsPBGhhyPoPLHZAuoZiiHo9q6FLlEeIDEXuj6T4=",
-        "owner": "divnix",
-        "repo": "blank",
-        "rev": "5a5d2684073d9f563072ed07c871d577a6c614a8",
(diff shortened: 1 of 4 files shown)


jberthold:
We should bump the version of K used by CI in the integration tests (nix) - I could not find the place where the version is stored, though.

goodlyrottenapple (commented):
LGTM, will look into the K version in CI. there might be a bit of tweaking necessary since we now depend on booster in K, so there is a circular dependency, but I think it can be dealt with

geo2a (approved):

jberthold:
> LGTM, will look into the K version in CI. there might be a bit of tweaking necessary since we now depend on booster in K, so there is a circular dependency, but I think it can be dealt with

It should not be too distinct from the circular dependency between k and haskell-backend... we can probably use the same derivation as over there.

jberthold:
Merging this to unblock others - will try to follow up with the nix changes tomorrow.

PR #197, opened 2023-06-19
House-keeping: nix-action, SourceRef, simpler add-module test 

jberthold:
* upgrades the `install-nix-action` to `v22` (to avoid a MacOS permission issue in version 12.6.6)
* Uses `SourceRef` for definition summaries (large diff, sorry)
* updates `add-module` test to use a simpler rule format (no alias, no anti-left)

Best reviewed one commit at a time.


.github/workflows/test.yml
```diff
@@ -198,14 +198,12 @@ jobs:
           nix build .#hs-backend-booster:exe:rpc-client --extra-experimental-features 'nix-command flakes' --print-build-logs
 
       - name: Run unit tests
-        if: ${{ matrix.os != 'macos-12' }}
```
goodlyrottenapple, line 201: not sure its worth enabling these on the public runners? we already test on the M1 which is a lot faster

geo2a (approved):

.github/workflows/test.yml
jberthold: I was just wondering why we aren't running the tests. If there is a good reason not to we can revert that part.

jberthold: As discussed, integration tests take very long on the public runner - those are left disabled now.

PR #202, opened 2023-06-20
199 pass `get-model` through to kore-rpc

### LINKED ISSUE #199: Add `get-model` endpoint (pass-through) to booster proxy
The new `get-model` endpoint [has been implemented in `kore-rpc`](https://github.com/runtimeverification/haskell-backend/issues/3599), so `kore-rpc-booster` should have a pass-through implementation for it, too.

jberthold:
Fixes #199 

Also added `get-model` and `add-module` support to `rpc-client` and replicated the `get-model` integration tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -200,7 +200,8 @@ jobs:
       - name: Run unit tests
         env:
           GC_DONT_GC: '1'
-        run: nix build .#checks.unit-tests --extra-experimental-features 'nix-command flakes' --print-build-logs
+        run: nix build .#hs-backend-booster:test:unit-tests --extra-experimental-features 'nix-command flakes' --print-build-logs
+
 
       - name: Run integration tests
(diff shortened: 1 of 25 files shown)


.github/workflows/test.yml
```diff
@@ -200,8 +200,7 @@ jobs:
       - name: Run unit tests
         env:
           GC_DONT_GC: '1'
-        run: nix build .#hs-backend-booster:test:unit-tests --extra-experimental-features 'nix-command flakes' --print-build-logs
-
+        run: nix build .#checks.unit-tests --extra-experimental-features 'nix-command flakes' --print-build-logs
```
jberthold: While inspecting the integration test log output, I found that we are not actually _running_ the unit tests, we are just _building them_. However, this invocation clearly does not run them either, it fails.

geo2a (approved):

PR #203, opened 2023-06-20
rpc-client: refactor argument parser

geo2a:
To implement the `simplify-implication` (and `implies`) endpoint, we need to be able to pass two JSON files with Kore terms to `rpc-client`. This changes will allow to do that easily by refactoring the request as "commands" instead of "options", since "options" only allow up to one argument.


jberthold (approved):
LGTM

jberthold:
Going to auto-merge this now, has been sitting for 2 days.

PR #208, opened 2023-06-26
Simplify constraints in patterns

### LINKED ISSUE #206: `simplify` endpoint should also simplify constraints when given a constrained term
[Originally from slack by @geo2a](https://runtimeverification.slack.com/archives/C05CLPW6U05/p1687444730987109)
The `simplify` endpoint currently only simplifies the term but not the constraints when given a constrained term. Both term and constraints (one at a time for now) should be simplified.

geo2a:
Closes #206

* add `simplifyPattern` that simplifies both the pattern's term and it's constraints
* add somewhat bogus unit tests that duplicate the ones for `evaluateTerm`, but operate on `Pattern`. Proper constraint simplification needs the LLVM backend and thus is not done with unit tests.


goodlyrottenapple (approved):
LGTM

PR #214, opened 2023-06-29
Build systems maintanence 

geo2a:
* Bump GHC to 9.2.8 in the Nix flake
* Bump GHC tp 9.2.8 in CI action
* Bump Stack resolver to lts-20.26
* Freeze cabal dependencies


jberthold:
What is actually needed is to downgrade `aeson`, not `yaml` (`yaml` is missing an upper bound on the `aeson` dependency). I did not push that change because our `MacM1` runner was out of order until just now...

We could fix this PR as described and merge it (after merging the `haskell-backend` sibling PR) but my other PR #215 is probably ready to go, so we could follow up with the full upgrade after merging that one.

jberthold (approved):
LGTM

PR #220, opened 2023-07-04
Fix Nix flake development shell

geo2a:
It turns out we've indeed broke the development shell (specifically, the build of `fourmolu`) of the Nix flake by updating the Hackage index state. I've specified the old index state specifically for `fourmolu`. I would be better to replicate the [build command](https://github.com/runtimeverification/hs-backend-booster/blob/b76b4fd04dd93407b4d6f6606ba763ac36fe4f80/.github/workflows/Dockerfile#L53) for `fourmolu` from our Dockerfile that constraints `aeson`, but I could not figure out how to do that with `haskell.nix`.

* use older index state for fourmoulu
* Run `nix develop` in CI to check that the shell builds


jberthold (approved):
LGTM

Issue #126, opened 2023-02-17
Make Kore.Pattern.Rewrite monad stack IO based


goodlyrottenapple:
The new tracing code in #122  introduces a timing function which emits a `START <id>/STOP <id>` message into the event log around a block of IO code. This can be used to produce timing data and a speedscope profile. Unfortunately, I couldn't get this to work in pure/monad only context, due to Haskell's evaluation of non-IO code, as `eventlog-parser` is based on `eventlog2speedscope`, which requires that the `START/STOP` messages are ordered/nested correctly. This isn't guaranteed in non-IO code.

One option to better instrument the re-writing/unification code is to make the monads into IO, which would guarantee the execution order. @ana-pantilie @jberthold if you have ideas about how to implement this kind of bracketing in pure code reliably, let me know...

PR #225, opened 2023-07-06
Don't trace rewrites and equation applications if not requested

### LINKED ISSUE #224: Don't collect equation traces when not asked for
Profiling an ERC20 bug report revealed that we allocate a lot of memory for equation traces, even if we don't send them back or log them.

We need to thread the configuration from the request to the place where we trace equation and don't trace them if not needed.

geo2a:
Fixes #224 and #126

* Refactor `EquatioM` to `EquationT`
  - make it a monad transformer
  - Move immutable data to a `ReaderT`
  - Add `MonadLoggerIO` constraint on interface functions
* Refactor `RewriteM` to `RewriteT`
  - make it a monad transformer
  - Make the reader config a proper product type, rather then a tuple
  - Add `MonadLoggerIO` constraint on interface functions
* Thread tracing flag down to rewrites and equation applications. Only accumulate traces if the flag is set.

Currently, the `doTracing` flag is less granular than the RPC request tracing options: we will collect all traces if any of the four RPC options is set. We should probably address this in a follow-up PR.

### DIFF
--- a/library/Booster/Pattern/ApplyEquations.hs
+++ b/library/Booster/Pattern/ApplyEquations.hs
@@ -14,7 +14,6 @@ module Booster.Pattern.ApplyEquations (
     isMatchFailure,
     isSuccess,
     simplifyConstraint,
-    traceSimplifyConstraint,
 ) where
 
 import Control.Monad
@@ -530,18 +529,8 @@ simplifyConstraint ::
     KoreDefinition ->
     Maybe LLVM.API ->
     Predicate ->
-    Predicate
-simplifyConstraint doTracing def mbApi p =
-    either (const p) fst $ traceSimplifyConstraint doTracing def mbApi p
-
--- | Constraint simplification that collects a simplification trace
-traceSimplifyConstraint ::
-    Bool ->
-    KoreDefinition ->
-    Maybe LLVM.API ->
-    Predicate ->
     Either EquationFailure (Predicate, [EquationTrace])
-traceSimplifyConstraint doTracing def mbApi p =
+simplifyConstraint doTracing def mbApi p =
     runEquationM def mbApi doTracing $ simplifyConstraint' p
 
 -- version for internal nested evaluation
--- a/library/Booster/Pattern/ApplyEquations.hs
+++ b/library/Booster/Pattern/ApplyEquations.hs
@@ -170,13 +170,15 @@ data EquationPreference = PreferFunctions | PreferSimplifications
     deriving stock (Eq, Show)
 
 runEquationM ::
+    Bool ->
     KoreDefinition ->
     Maybe LLVM.API ->
-    Bool ->
     EquationM a ->
     Either EquationFailure (a, [EquationTrace])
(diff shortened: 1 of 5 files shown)


library/Booster/Pattern/Rewrite.hs
```diff
@@ -425,12 +437,12 @@ performRewrite def mLlvmLibrary mbMaxDepth cutLabels terminalLabels pat = do
...
     rewriteTrace t = do
         logRewrite $ pack $ renderDefault $ pretty t
-        modify $ \(counter, traces) -> (counter, traces |> t)
+        when doTracing $ modify $ \(counter, traces) -> (counter, traces |> t)
     incrementCounter = modify $ \(counter, traces) -> (counter + 1, traces)
 
     simplifyP :: Pattern -> StateT (Natural, Seq (RewriteTrace Pattern)) io Pattern
     simplifyP p = do
-        let result = evaluateTerm TopDown def mLlvmLibrary p.term
+        let result = evaluateTerm doTracing TopDown def mLlvmLibrary p.term
```
jberthold, line 445: We are losing the `-l Simplify` log if we do this... (see `logTraces` below).
Somewhat lame that one cannot find out whether that log level is active or not...
That problem goes away if we make `EquationM` a `MonadLoggerIO` instance (but that might be a bigger change)

jberthold (dismissed):
We should maybe merge this as a stop-gap right now, but immediately follow up with a fix to get back `--log-level Simplify` (on a new ticket, making an `instance MonadIO EquationM` and emitting `Simplify` logs directly instead of storing them).

jberthold (commented):
Thanks for picking this up quickly. However, I am afraid this cannot be merged as-is now, we need to rearrange the calls to `run*LoggingT`

library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -163,12 +181,19 @@ data EquationPreference = PreferFunctions | PreferSimplifications
...
     KoreDefinition ->
     Maybe LLVM.API ->
     EquationM a ->
-    Either EquationFailure (a, [EquationTrace])
-runEquationM definition llvmApi (EquationM m) =
-    fmap (fmap $ toList . trace) <$> runExcept $ runStateT m $ startState definition llvmApi
+    IO (Either EquationFailure (a, [EquationTrace]))
+runEquationM doTracing definition llvmApi (EquationM m) = do
+    endState <-
+        runStdoutLoggingT
```
jberthold, line 191: This call to `runStdoutLoggingT` doubles up the logging - the `respond` function in `Booster.JsonRpc` itself already runs in an `m` with  `MonadLoggerIO m` context.
The `runEquationM` type should be `MonadLoggerIO m => ... m (Either ...)`, otherwise we need to control the logging (filters etc) in two places to avoid discrepancies. One of which is that [we actually log to `stderr` in the server](https://github.com/runtimeverification/hs-backend-booster/blob/main/tools/booster/Server.hs#L94).

PR #227, opened 2023-07-11
200 - Collect constraints in equation state, abort execution on `ensures(false)`

### LINKED ISSUE #200: Retain `ensures` clauses from simplifcation and function equations
The `ensures` clauses in an equation are currently not added to the state when applying an equation. They should be added to retain the knowledge about the state.

From initial code exploration:

Code in `Pattern.Rewrite` does the right thing: prune any `ensures(\bottom)` branches and filter out `ensures(\top)`.

When applying equations (in `Pattern.ApplyEquation`, we do not consider the `ensures` so far.
* What should be done:
  - remove all ensures(true), known to be true (trivial)
  - retain all ensures(X) that are undetermined
* What about ensures(false) in equations?
    - skipping (`continue`) for functions means we leave the function unevaluated (function should be deterministic)
    - similar for simplifications, but other simplifications might apply (if the semantics is weird/non-deterministic).
    - failing (`abort`) means to _return the original term_ (without indication that it was evaluating/simplifying to `\bottom`)
    - We could also retain the result that is known to be bottom (i.e., consider this a success). This bottom won't be looked-at      further in subsequent rewrites - wasted work.
    - instead, we should _prune_ the entire execution here
    - this is a new code path that we don't have so far.
      - we could just return a plain bottom if it was a term, but it is not
      - we can directly return the last known config and a plain bottom predicate (from the rewriter or the simplifier)

jberthold:
Fixes #200 

* `simplifyPattern` has changed to `evaluatePattern` to avoid confusion with the LLVM-driven `simplifyPattern`.
* Side conditions (`constraints` in a pattern) are stored during evaluation of function and simplification rules. When evaluating/simplifying a predicate, it is temporarily removed from the stored side conditions. (This prepares for adding an SMT solver and checking constraints in combination later).
* `ensures(P)` clauses in equations add to the stored side conditions, the returned pattern contains these new side conditions as well as the (simplified) original ones. 
* When an `ensures` clause is found to be (trivially) false, the evaluation is pruned.
* This makes the rewrite `stuck` if the simplification is called from `performRewrite`.
* Simplification of rewrite results before they are returned may change the result  because parts may be stuck. In case of branching, the execution may continue if all branches but one are pruned, other results may turn into `stuck`.


library/Booster/JsonRpc.hs
```diff
@@ -509,4 +516,11 @@ mkLogRewriteTrace
                                     , origin = Booster
                                     , result = Failure{reason = "Internal error: " <> err, _ruleId = Nothing}
                                     }
+                            ApplyEquations.SideConditionsFalse{} ->
+                                Simplification -- FIXME
```
geo2a, line 520: What this FIXME refers to?

library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -242,6 +258,43 @@ evaluateTerm' ::
...
+    MonadLoggerIO io =>
+    Pattern ->
+    EquationT io Pattern
+evaluatePattern' Pattern{term, constraints} = do
+    pushConstraints constraints
+    newTerm <- evaluateTerm' TopDown term
+    -- after evaluating the term, evaluate all (existing and
+    -- newly-acquired) constraints, once
+    currentPredicates <- Set.toList . predicates <$> getState
+    traverse_ simplifyAssumedPredicate currentPredicates
```
geo2a, line 285: Can we use `mapM_` instead of `traverse_` here to avoid converting to the set to list? I.e.

```suggestion
    currentPredicates <- predicates <$> getState
    mapM_ simplifyAssumedPredicate currentPredicates
```

geo2a: Or actually, we could just use `traverse_` and not convert to list. Anyway, this is just a nitpick, feel free to ignore.

library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -465,24 +521,32 @@ applyEquation term rule = fmap (either id Success) $ runExceptT $ do
...
+                                    concatMap (splitBoolPredicates . substituteInPredicate subst) $
+                                        rule.ensures
+                            mbEnsuredConditions <-
+                                runMaybeT $ catMaybes <$> mapM checkConstraint ensured
+                            maybe
+                                (throwE $ EnsuresFalse ensured)
+                                -- throws if an ensured condition found to be false
+                                (lift . pushConstraints)
+                                -- pushes new ensured conditions and return result
+                                mbEnsuredConditions
```
geo2a, line 548: Is there a good reason to not do explicit pattern-matching on `mbEnsuredConditions`? I think `maybe` is generally harder to read than an explicit case statement.

```suggestion
                            case mbEnsuredConditions of
                                -- throws if an ensured condition found to be false
                                Nothing -> throwE $ EnsuresFalse ensured
                                -- pushes new ensured conditions and return result
                                Just conditions -> lift . pushConstraints $ conditions
```

geo2a (approved):
I've left a couple of nitpicks, but nothing significant, so I'm approving!

library/Booster/JsonRpc.hs
jberthold: Yeah, maybe acceptable that the field `originalTerm` is not provided here.
Ah... and now I also remember I was going to add simplification traces to the error constructor...

library/Booster/Pattern/ApplyEquations.hs
jberthold: Good point, using `traverse_` directly then.

library/Booster/JsonRpc.hs
```diff
@@ -509,4 +516,11 @@ mkLogRewriteTrace
                                     , origin = Booster
                                     , result = Failure{reason = "Internal error: " <> err, _ruleId = Nothing}
                                     }
+                            ApplyEquations.SideConditionsFalse _predicates _traces ->
+                                Simplification
+                                    { originalTerm = Nothing
```
jberthold, line 521: While removing the `FIXME` that I had here, I noticed that we are actually losing traces when we hit an error after a number of successful simplifications. For `SideConditionsFalse` as well as `EquationLoop`, we would in principle have a sequence of traces that we could return to the caller for interpretation.

Issue #229, opened 2023-07-11
Rename `dltest` executable


Baltoli:
The name of this binary clashes with something in our transitive Homebrew dependency chain when building K on macOS: https://github.com/runtimeverification/k/actions/runs/5520367793/jobs/10066915489

Can this binary be renamed to something similar, or not installed if we don't need it to run the booster as part of K?

PR #230, opened 2023-07-11
Remove `dltest`

goodlyrottenapple:
The `dltest` binary was an early teting tool for integrating the LLVM backend API. It has since been replaced with a proper integration test and is now causing some issues #229 so this PR removes it completely.


Baltoli (approved):

jberthold (approved):
:fire:

PR #232, opened 2023-07-12
231 Refactor match1, modifying variable and symbol cases. Adapt tests

### LINKED ISSUE #231: Problems with matching algorithm (variables, injections)
The current matching algorithm has several problems:

* [Matching a `\dv{...}(...)` with a variable (in the subject) _fails_](https://github.com/runtimeverification/hs-backend-booster/blob/main/library/Booster/Pattern/Match.hs#L144-L148) but should be indeterminate
* Generally, variables in the subject should produce an _indeterminate_ match result unless they are matched by a variable in the pattern.

Several unit tests will have to be revised because they use variables as placeholders.

* [Matching an injection with a constructor symbol](https://github.com/runtimeverification/hs-backend-booster/blob/main/library/Booster/Pattern/Match.hs#L181-L185) should fail ~recurse into the subterm if the sorts match.~

While fixing this, the whole `match1` function should be divided into sub-cases for better overview.

jberthold:
Fixes #231 
* Variables on the RHS of a match produce an indeterminate result
  unless LHS is also a variable
* Previously there were a few cases where all symbol applications
  produced indeterminate results but only functions should.
* Tests that previously used variables as function arguments had to be
  adapted (by wrapping the variable in a constructor, to retain
  symbolicness of the function argument)
* The `match1` workhorse function of the matcher was refactored and
  commented for better overview

### DIFF
--- a/library/Booster/Pattern/Match.hs
+++ b/library/Booster/Pattern/Match.hs
@@ -23,6 +23,7 @@ import Data.Sequence (Seq (..), (><))
 import Data.Sequence qualified as Seq
 import Data.Set (Set)
 import Data.Set qualified as Set
+import Prettyprinter
 
 import Booster.Definition.Base
 import Booster.Pattern.Base
@@ -31,6 +32,7 @@ import Booster.Pattern.Util (
     checkSymbolIsAc,
     freeVariables,
     isConstructorSymbol,
+    isFunctionSymbol,
     modifyVariablesInP,
     sortOfTerm,
     substituteInTerm,
@@ -56,6 +58,13 @@ data MatchFailReason
     | -- | The two terms have differing argument lengths
       ArgLengthsDiffer Term Term
     deriving stock (Eq, Show)
+
+instance Pretty MatchFailReason where
+    pretty = \case
+        General reason -> pretty reason
+        SharedVariables vs -> "Shared variables:" <+> hsep (map pretty $ Set.toList vs)
+        SubsortingError err -> pretty $ show err
+        ArgLengthsDiffer t1 t2 -> vsep ["Argument length differ", pretty t1, pretty t2]
 
 {- | Attempts to find a matching substitution for the given
    term1 to term2.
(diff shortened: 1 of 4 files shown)


unit-tests/Test/Booster/Pattern/ApplyEquations.hs
```diff
@@ -36,29 +36,37 @@ test_evaluateFunction =
         "Evaluating functions using rules without side conditions"
         [ -- f1(a) => a
           testCase "Simple function evaluation" $ do
-            eval TopDown [trm| f1{}(A:SomeSort{}) |] @?= Right [trm| A:SomeSort{} |]
-            eval BottomUp [trm| f1{}(A:SomeSort{}) |] @?= Right [trm| A:SomeSort{} |]
-        , -- f2(f1(f1(a))) => f2(a). f2 is marked as partial, so not evaluating
+            eval TopDown [trm| f1{}(con2{}(A:SomeSort{})) |] @?= Right [trm| con2{}(A:SomeSort{}) |]
+            eval BottomUp [trm| f1{}(con2{}(A:SomeSort{})) |] @?= Right [trm| con2{}(A:SomeSort{}) |]
+        , -- f2(f1(f1(con2(a)))) => f2(con2(a)). f2 is marked as partial, so not evaluating
```
goodlyrottenapple, line 41: i would have thought these should be indeterminate...
shouldn't any term with a variable return an indeterminate match??

goodlyrottenapple: e.g. given

```
rule f(foo(bar())) => baz()
rule f(foo(A)) => A
```

where `foo`,`bar` and `baz` are constructors, we don't know what `f(foo(X))` should rewrite to

jberthold: That's kind-of what we had here, but it is not _any_ variable occurrence, just the variable being at the top matching the entire argument should be indeterminate.
The test fixture has rules 
```
f ( con1 ( X ) )                  => con2 ( f1 ( X ) )
f ( inj{..} ( con4 ( X , _Y ) ) ) => X
f ( X )                           => X
``` 
Previously  the test was using `f1 ( A )` (should be indeterminate indeed), now we use `f1 (con2 ( A ))`

library/Booster/Pattern/Match.hs
```diff
@@ -126,26 +135,31 @@ match1
                     then term2
                     else Injection termSort variableSort term2
                 )
+-- subject term is a variable but pattern term is not: indeterminate
+match1
+    pat
+    var@Var{} =
+        indeterminate pat var
```
jberthold, line 142:
```suggestion
        indeterminate pat var
-- and-terms in subject term are considered indeterminate 
-- (what would they mean?)
match1
    pat
    and@And{} =
        indeterminate pat and
```

geo2a (approved):

PR #234, opened 2023-07-13
Connect auto update to existing update script

jberthold:
Related to #233 , extending the update job scope

* Update `cabal.project` and `stack.yaml` with the desired commit hash and correct sha (uses existing script)
* Update `cabal.project.freeze` (in case the updated haskell-backend adds/removes library dependencies)
  This uses another existing script, called from the first one (changed).
* Do this only if the `flake.nix` file has actually changed (to avoid running `nix flake update` without need)
* Combine all changes in a single commit for better overview.

Also contains other small clean-ups in `flake.nix` and `cabal.project.freeze`.


jberthold:
I tested the changes to `update-haskell-backend.sh` and `freeze-cabal-to-stack-resolver.sh` by running the update script - `haskell-backend` is one commit ahead of what is currently checked in.

scripts/freeze-cabal-to-stack-resolver.sh
```diff
@@ -2,8 +2,15 @@
...
+# must remove the prior freeze file to actually update
+rm -f cabal.project.freeze
+
 stack --test ls dependencies \
     | sed -e 's/^\([a-zA-Z0-9.-]*\) \([0-9.]*\)/--constraint="\1 == \2"/' \
     | xargs -x cabal freeze --enable-tests
 
+# ghc-bignum is removed from the freeze file because it is wired into
+# GHC and creates problems when nix builds tools with different GHCs
+sed -i -e '/any\.ghc-bignum/d' cabal.project.freeze
```
goodlyrottenapple, line 14: we should add alex and happy here as well, because there never seems to be the exact version of these packages in the nix version of cabal

goodlyrottenapple (approved):

Issue #207, opened 2023-06-26
Ability to consume RPC bug report tarballs


jberthold:
Pyk can generate a tarball of all RPC interactions (as raw unformatted json), which can be used to replay execution sequences. 

We should use this as a regression test tool.

While we cannot expect verbatim identical json output, the delta should be limited to known changed behaviours.

This will require a json-diff to work efficiently on it
  - possibly also store formatted json for an easier git diff
  - can [try to use `aeson-diff`](https://hackage.haskell.org/package/aeson-diff) instead of formatting the json as a preprocessing step

* tool to extract the RPC files from a tarball (can use `run-ghc`, no need for a full CLI app)
  - formats all json files for better `git diff`
* test harness to execute all RPC interactions from a directory (in order) and compare response files

These could possibly be run from the same code.

jberthold:
Steps towards running bug report tarballs as regression tests:
* [X] Modify the `bug_report.tar` to a) include information about the LLVM interpreter in case `booster` was used, and b) remove unnecessary large files from the definition 
* [ ] add tool code to booster which categorises kore-rpc JSON files and displays differences between them
* [ ] extend `rpc-client` to unpack a tarball and run all requests in `rpc_*` inside it, comparing results
  - should be configurable to stop on the first error or continue
  - should expect a running server ready to receive requests,
  - and display the command line to start one (from information stored in the tarball) if no server is found

PR #238, opened 2023-07-20
207 kore rpc json differ

jberthold:
Part of #207 

* LIbrary module to classify json files into the types used for `kore-rpc-types` and compute differences between files of identical type
* Small executable to load two files and determine the differences between them
* Some tests with exemplary json files have been added (comparing pair-wise, with golden files for expected output) 

Also in this PR:

*  Add dry-run mode to `rpc-client`
   * A new switch `--dry-run` makes `rpc-client` print the request instead of sending it
   * The option can be combined with `--output/-o` and `--prettify`


geo2a (approved):
Looks good to me! I have not tried it out yet, but the test suite seems comprehensive enough to merge. Thanks @jberthold!

Issue #236, opened 2023-07-14
Return traces to caller when `EquationLoop` or `SideConditionsFalse` 


jberthold:
_Originally posted by @jberthold in https://github.com/runtimeverification/hs-backend-booster/pull/227#discussion_r1263268989_

Looking at `Booster.JsonRpc.mkLogRewriteTrace`, I noticed that we are actually losing traces when we hit an error after a number of successful simplifications. For `SideConditionsFalse` as well as `EquationLoop`, we would in principle have a sequence of traces that we could return to the caller for interpretation.

PR #240, opened 2023-07-24
Reshuffle the EquationT transformer stack

goodlyrottenapple:
Fixes https://github.com/runtimeverification/hs-backend-booster/issues/236 by reshuffling the `EquationT` transformer, currently defined as

```Haskell
newtype EquationT io a
    = EquationT (StateT EquationState (ReaderT EquationConfig (ExceptT EquationFailure io)) a)
 -- ~ EquationState -> EquationConfig -> io (Either EquationFailure (a, EquationState))
```

into 

```Haskell
newtype EquationT io a
    = EquationT (ReaderT EquationConfig (ExceptT EquationFailure (StateT EquationState io)) a)
 -- ~ EquationConfig -> EquationState -> io (Either EquationFailure a, EquationState)
```

This ensures that we always return the state even after a failed computation, hence we don't need to include the traces as data within the `EquationFailure`

### DIFF
--- a/library/Booster/JsonRpc.hs
+++ b/library/Booster/JsonRpc.hs
@@ -144,7 +144,7 @@ respond stateVar =
                 Right (TermAndPredicate pat) -> do
                     Log.logInfoNS "booster" "Simplifying a pattern"
                     ApplyEquations.evaluatePattern doTracing def mLlvmLibrary pat >>= \case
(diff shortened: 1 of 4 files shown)


jberthold (approved):
LGTM. Nice one :sparkles:

PR #241, opened 2023-07-25
207 consume pyk rpc bug reports

jberthold:
Extend `rpc-client` to run all requests from a tarball (from the directory `rpc_*` inside it)

Also refactored program option data type(s) to accommodate the new run mode.


jberthold:
I tried to add an integration test with a foundry bug report, but pushed it to a separate branch (https://github.com/runtimeverification/hs-backend-booster/compare/207-consume-pyk-rpc-bug-reports...tmp/207-add-regression-test-with-foundry-bugreport).
Unfortunately the test quickly fails because of missing krypto-plugin functions, like so:
```
[Info] Process request 2 simplify
[Info#booster] Simplifying a pattern
../../.build/booster/bin/kore-rpc-booster: symbol lookup error: ./resources/foundry-bug-report.tar.gz.dylib: undefined symbol: hook_KRYPTO_keccak256
[Info] Round trip time for request '002' was 19.163389 ms
[Info] Response received.
[Error] Request 002 failed: Json data in files is of different type
  * File expected response: SimplifyM response
  * File actual response: non-JSON file
```
Guess the regression tests are still a bit further out then...

test/jsonrpc-examples/add-module.error.response
```diff
@@ -0,0 +1 @@
+{"jsonrpc":"2.0","id":4,"error":{"code":2,"data":{"context":null,"error":"FOUNDRY-MAIN-DEPENDS-MODULE: Duplicate module"},"message":"Could not verify pattern"}}
```
jberthold, line 1: This was a missing test case, and it hit a pattern match error in the `decodeKoreRpc` function

tools/rpc-client/RpcClient.hs
```diff
@@ -22,81 +27,125 @@ import Data.Aeson.KeyMap qualified as JsonKeyMap
...
+    Options{common, runOptions} <- execParser parseOptions
+    case runOptions of
+        RunTarball tarFile keepGoing ->
+            runTarball common.host common.port tarFile keepGoing
+        RunSingle mode optionFile options processingOptions -> do
+            let ProcessingOptions{postProcessing, prettify, time, dryRun} = processingOptions
+
+            request <-
+                trace "[Info] Preparing request data" $
+                    prepareRequestData mode optionFile options
```
jberthold, line 65: This is still using `trace` throughout but we should convert it to using a logger (just `MonadLoggerIO` like the server, I would guess)
Also, we should send a `Cancel` for the request in-flight when the program is interrupted (`^C`).

I left those two changes for a follow-up PR, as this one is certainly enough change for one round of review.

tools/rpc-client/RpcClient.hs
```diff
@@ -202,38 +244,206 @@ parsePostProcessing =
...
+                , "2) look up the module name `<MODULE>` in `server_instance.json`;"
+                , "3) and then run the server using"
+                , "   $ kore-rpc definition.kore --main-module <MODULE> --server-port " <> show port
+                , ""
+                , "If you want to use `kore-rpc-booster, you should als compile an LLVM backend library"
+                , "by 1) extracting the `llvm_definition/` directory from the tarball;"
+                , "   2) running the llvm backend compiler on the unpacked files:"
+                , "      $ llvm-kompile llvm_definition/definition.kore llvm_definition/dt c`"
+                , "This will generate `interpreter.so` and you can run"
+                , "  `kore-rpc-booster definition.kore --main-module <MODULE> --llvm-backend-library interpreter.so`"
```
goodlyrottenapple, line 421: this is great!!

goodlyrottenapple (approved):

tools/rpc-client/RpcClient.hs
geo2a: I second this: I was hot-testing these features on a bug report, and this was very helpful!

geo2a (approved):

tools/rpc-client/RpcClient.hs
```diff
@@ -202,38 +244,206 @@ parsePostProcessing =
...
+                    pure mbError
+            shutdown skt ShutdownReceive
+            exitWith (if all isNothing results then ExitSuccess else ExitFailure 2)
+
+    -- complain on any errors in the tarball
+    throwAnyError :: Either Tar.FormatError Tar.FileNameError -> IO a
+    throwAnyError = either throwIO throwIO
+
+    -- unpack all rpc_*/*.json files into dir and return their names
+    unpackIfRpc :: FilePath -> Tar.Entry -> IO [FilePath] -> IO [FilePath]
```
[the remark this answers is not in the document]
jberthold: This is interesting... I think we can maybe deal with the current bug reports:
* create  the `rpc_*` directories (and only these) and unpack the json files into these directories
* return the full relative paths 
This should mean we will run all requests from all directories. Will try that tomorrow.

tools/rpc-client/RpcClient.hs
```diff
@@ -202,38 +244,206 @@ parsePostProcessing =
...
+                )
+            <> command
+                "run-tarball"
+                ( info
+                    ( RunTarball
+                        <$> strArgument (metavar "FILENAME")
+                        <*> switch (long "keep-going" <> help "do not stop on unexpected output")
+                        <**> helper
+                    )
                     (progDesc "check satisfiability/provide model for the state in the file")
```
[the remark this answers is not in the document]
jberthold: :man_facepalming: :paintbrush:

PR #242, opened 2023-07-26
207 test tarball runner + tweak rpc client logging

### LINKED ISSUE #207: Ability to consume RPC bug report tarballs

jberthold:
- [X] Tweaks to `rpc-client`:
  * use `MonadLogger` and a log-level option
  * cancel the request in flight if the user interrupts the program (by `^C` or sending `SIGINT`)
- [x] unpack and use all `rpc_*` directories, with their respective path
- [x] Add a regression test for running a tarball.
 
Fixes #207


test/rpc-integration/resources/foundry-bug-report.tar.gz.kompile
```diff
@@ -0,0 +1,42 @@
...
+tar xzf ../test-$NAMETGZ -C $NAME definition.kore llvm_definition/
+
+# provide definition
+cp $NAME/definition.kore ${NAMETGZ}.kore
+
+# kompile llvm-definition to interpreter
+llvm-kompile $NAME/llvm_definition/definition.kore $NAME/llvm_definition/dt c -- \
+             -fPIC -std=c++17 -o interpreter \
+             $PLUGIN_LIBS $PLUGIN_INCLUDE $PLUGIN_CPP \
+             -lcrypto -lssl -lprocps $NIX_LIBS
```
jberthold, line 38: This script follows how we link plugin code to the llvm interpreter in `evm-semantics`:
- statically link in the plugin dependencies (`PLUGIN_LIBS`)
- compile (some of) the plugin's `*.cpp` files (`PLUGIN_CPP`), requiring the `include` directories of its dependencies (`PLUGIN_INCLUDE`)
- link (dynamically) to `crypto`, `ssl`, and `procps` (which is BTW Linux specific)
- access `NIX_LIBS` variable to obtain include and library directories for the latter

This has a couple of issues when run with `nix`:
1. the plugin dependencies are not actually built so the `LIBFILE` variable stays empty
2. there is no `NIX_LIBS` at the moment (would probably be easy enough to add?)

flake.nix
```diff
@@ -8,14 +8,18 @@
     haskell-nix.follows = "haskell-backend/haskell-nix";
```
jberthold, line 8: I noticed that we are testing with a 5.6 version of K here... but upgrading seems to cause issues with the index-state being too new so I left that for a follow-up.

test/rpc-integration/default.nix
```diff
@@ -16,7 +16,7 @@ let
           export ${f}
         '') buildFlags}
         export SERVER=${kore-rpc-booster}/bin/kore-rpc-booster
-        export CLIENT=${rpc-client}/bin/rpc-client
+        export CLIENT="${rpc-client}/bin/rpc-client -l debug"
```
jberthold: I ran with debug logging to analyse the M1 failure but it is flaky... passed on the 2nd run after truncating some output on the first run.
Hopefully this won't become a problem later if/when we merge this.

goodlyrottenapple (approved):

PR #245, opened 2023-08-01
Misc fixes

goodlyrottenapple:
* Add guard to check the old backend has made at least one step in the proxy server
*  Return the indeterminate conditions when throwing IndeterminateCondition for more informative trace

### DIFF
--- a/tools/booster/Proxy.hs
+++ b/tools/booster/Proxy.hs
@@ -179,7 +179,12 @@ respondEither mbStatsVar booster kore req = case req of
                                 -- if we made one step, add the number of
                                 -- steps we have taken to the counter and
                                 -- attempt with booster again
-                                Log.logInfoNS "proxy" "kore depth-bound, continuing"
+                                when (koreResult.depth == 0) $ error "Expected kore-rpc to take at least one step"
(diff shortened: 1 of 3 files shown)


geo2a (approved):

tools/booster/Proxy.hs
```diff
@@ -179,7 +179,12 @@ respondEither mbStatsVar booster kore req = case req of
                                 -- if we made one step, add the number of
                                 -- steps we have taken to the counter and
                                 -- attempt with booster again
-                                Log.logInfoNS "proxy" "kore depth-bound, continuing"
+                                when (koreResult.depth == 0) $ error "Expected kore-rpc to take at least one step"
```
jberthold, line 182: Are we _sure_ that in all `aborted` cases `kore-rpc` is able to take a step?
(Maybe `booster` is unable to detect that the execution is stuck, is that possible?)
What should definitely not happen is `reason = depth-bound` _and_ `depth == 0`

jberthold (approved):
LGTM. We need to also fix the actual problem in  `kore-rpc` , of course...

PR #247, opened 2023-08-01
Add `--hijack-eventlog-file` to emit `LlvmCall`s into a file instead of GHC eventlog

geo2a:
This PR constitutes:
* a refactoring of `LlvmCall` event emitted to GHC eventlog;
* a possibility to force `LlvmCall` events to bypass the GHC eventlog and redirect them into a file. This becomes necessary if the calls are longer than 2^16 bytes, which happens for declaration of long EVM bytecode tokens.


geo2a:
I think it is useful to have the ability to trace calls in one form or another. I think we should keep it and perhaps we'll be able to reuse the infrastructure for something else LLVM-related later.

PR #249, opened 2023-08-01
Add an experimental option to simplify after executing

geo2a:
Revival of work done by @jberthold in #248

> * Option `--simplify-after-execute` runs the kore simplifier before returning the final result of an `execute` request
> * This can lead to inconsistent results due to simplifying result states to `\bottom`.
>   - When the result state simplifies to `\bottom`, the server returns `stuck` with this bottom state;
>   - When all `next-states` in a `cut-point-rule` or `branching` result simplify to `\bottom`, the server returns `stuck`;
>   - When only one of the branches in a `branching` result does not simplify to `\bottom`, the server returns the unsimplified result.
> 
> Rationale: Avoiding to return an inconsistent result is the most important requirement, and the client will most likely send back the state which is known to be bottom anyway. The `execute` request would have continued executing if we had pruned the branch earlier. We could try to continue (including the simplification step in the execute `loop`, but this is left for a subsequent pull request.


library/Booster/Syntax/Json/Externalise.hs
```diff
@@ -88,14 +88,14 @@ externalisePredicate sort =
...
-                    , varSort = externaliseSort $ var.variableSort
+                    , varSort = externaliseSort var.variableSort
                     , arg = recursion p1
                     }
             Internal.Forall var p1 ->
                 Syntax.KJForall
                     { sort
                     , var = varNameToId var.variableName
-                    , varSort = externaliseSort $ var.variableSort
+                    , varSort = externaliseSort var.variableSort
```
jberthold, line 98: Really saving a lot of dollars :moneybag: , great !!!  :smirk:

jberthold (approved):

library/Booster/Syntax/Json/Externalise.hs
geo2a: It's `hlint`, probably we should forward the dollars to them!

PR #251, opened 2023-08-02
Update stack.yaml.lock when auto-upgrading haskell-backend dependency

### LINKED ISSUE #250: mvn clean leaves booster submodule in dirty state
If you check out the frontend repo, run `mvn clean`, then run `git status`, you will see this:

```
On branch develop
Your branch is up to date with 'origin/develop'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
  (commit or discard the untracked or modified content in submodules)
        modified:   hs-backend-booster/src/main/native/hs-backend-booster (modified content)

no changes added to commit (use "git add" and/or "git commit -a")
```

If you cd into the submodule you can see that this is what was modified:

```
HEAD detached at a258fe8
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   stack.yaml.lock

no changes added to commit (use "git add" and/or "git commit -a")
```

We're running the same command in the haskell backend, but this doesn't happen. Whatever was done there to prevent this from happening needs to be done to the booster too.

jberthold:
Fixes #250

Also contains changes to stabilise the new tarball integration test.

### DIFF
--- a/.github/workflows/update-deps.yml
+++ b/.github/workflows/update-deps.yml
@@ -68,7 +68,7 @@ jobs:
           if ${changed}; then
             cabal update
             scripts/update-haskell-backend.sh ${haskell_backend_version}
-            git add flake.lock stack.yaml cabal.project cabal.project.freeze
+            git add flake.lock stack.yaml stack.yaml.lock cabal.project cabal.project.freeze
             git commit -m "update haskell-backend to version ${haskell_backend_version}"
             git push
           fi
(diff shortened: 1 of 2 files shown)


geo2a (approved):
LGTM, thanks Jost!

PR #253, opened 2023-08-04
DONOTMERGE Call KLLVM API to free memory

geo2a:
The K LLVM backend bindings provide the `kllvm_init` and `kllvm_free_all_memory` functions that initialize KLLVM memory andinvoke the KLLVM garbage collector. This PR is an attempt to call this function after calling `kore_simplify` from Booster.

Unfortunately, the results are unsatisfactory: initializing memory before simplification and calling GC after seems to change the behavior of `kore_simplify` in an unpredictable way. Specifically, when passed an empty byte array, the simplification returns a non-empty one (see this [unit-test](https://github.com/runtimeverification/hs-backend-booster/blob/97b28701430874e12af02bea40bf254c8519beff/test/llvm-integration/LLVM.hs#L80)). This behavior does not happen if we don't call `kllvm_init` and `kllvm_free_all_memory`.

Considering all this, I think we should de-prioritize investigating this leak and perhaps side-step it by periodically re-initializing the server in long-running proofs.

### DIFF
--- a/library/Booster/LLVM.hs
+++ b/library/Booster/LLVM.hs
@@ -2,7 +2,7 @@ module Booster.LLVM (simplifyBool, simplifyTerm) where
 
 import Control.Monad.IO.Class (MonadIO (..))
 import Data.Binary.Get
-import Data.ByteString (fromStrict)
+import Data.ByteString qualified as BS (fromStrict)
 import Data.Map qualified as Map
 import Data.Set qualified as Set
 import System.IO.Unsafe (unsafePerformIO)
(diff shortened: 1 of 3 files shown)


geo2a:
It turns out, that `kllvm_free_all_memory` is not what we need. Closing this until further developments on the LLVM backend side, see https://github.com/runtimeverification/llvm-backend/issues/816

goodlyrottenapple:
can we close this?

geo2a:
Yes, certainly! Closing as completed by #263.

PR #257, opened 2023-08-08
255 evaluate constraints recursively

### LINKED ISSUE #243: Simplifications not applied when running `foundry-prove` using the booster backend
When running a certain proof with `kevm foundry-prove` with the `--use-booster` option, I get a branching condition as follows, where `BUF` is a large `b"..."` expression (that I believe corresponds to some contract bytecode):

`{ true #Equals #asWord ( #range ( #buf ( 32 , ( maxUInt160 &Int #lookup ( ?STORAGE:Map , 3 ) ) ) +Bytes #buf ( 32 , #lookup ( ?STORAGE2:Map , 0 ) ) +Bytes #buf ( 32 , #lookup ( ?STORAGE3:Map , 0 ) ) +Bytes #buf ( 32 , #lookup ( ?STORAGE4:Map , 0 ) ) +Bytes b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00.#M\xaeu\xc7\x93\xf6z5\x08\x9c\x9d\x99$^\x1cXG\x0b" +Bytes #buf ( 32 , #lookup ( ?STORAGE0:Map , 4 ) ) +Bytes BUF , 160 , 32 ) ) <=Int maxUInt64 }
`

However, this expression could be simplified down to `#lookup ( ?STORAGE0:Map , 4 ) <=Int maxUInt64` by applying the following two lemmas, which already exist in `bytes-simplification.k`:
```
    rule [range-included-in-cHead]:
      #range(B1:Bytes +Bytes _:Bytes, S:Int, W:Int) => #range(B1, S, W)
      requires S +Int W <=Int lengthBytes(B1)
      [simplification]

    rule [range-outside-cHead]:
      #range(B1:Bytes +Bytes B2:Bytes, S:Int, W:Int) => #range(B2, S -Int lengthBytes(B1), W)
      requires lengthBytes(B1) <=Int S
      [simplification]
```
In fact, if I run the above expression as a `runLemma/doneLemma` claim, it does simplify as expected. Also, if the same proof is run using the legacy backend instead of the booster, the expression is also simplified. I don't know why the lemmas are not being applied when running the proof with the booster.

I have shared a bug report with @geo2a for a previous version of the proof that was exhibiting the same problem (but with a different expression). In both cases the same two lemmas above are not being applied.

### LINKED ISSUE #255: Simplify side conditions  while applying rules
As #243 turns out to be a major issue for real-world proofs, the booster backend should be extended to apply equation-based simplification (on symbolic terms) to conditions of rewrite rules.
Currently, conditions are only checked when they are fully concrete (i.e., they do not contain any variables) and booster aborts  rewriting if a rule condition cannot be determined. When applying function equations and a condition cannot be determined, evaluation stops (no further equations are attempted). When applying simplification equations, indeterminate conditions lead to skipping the equation in question.

- [X] Use simplification while checking symbolic conditions on rewrite rules. This was introduced together with #227 
- [x] Attempt to evaluate/simplify symbolic conditions on equations. This extension to the evaluator introduces recursive evaluation and needs to be protected against looping, the implementation is complex. It may also cause performance regressions.
- [x] For both cases, any indeterminate simplified conditions that are _ensured_ should be returned instead of the original ones

This work stream is an initial exploration of the potential of this change, a second iteration should follow whose design is informed by the initial findings.

jberthold:
Closes #255 
Fixes #243 

* When evaluating functions and applying simplifications, recursively simplify the side conditions of equations to test whether they can be applied.
* The recursion is bounded by the global iteration limit (currently 100 at all call sites), counting globally
* The original term stack is kept so that evaluation loops can be detected.
* If an error occurs in the recursive call, the function returns the original unevaluated constraint.


jberthold:
I tested this code with foundry tests in `evm-semantics`, there was no change against current mater, neither in the runtime  nor in the abort reasons :disappointed: - Still keen to merge it, though, since it solves the simplification problem reported in #243

test/rpc-integration/resources/simplify.k
```diff
@@ -1,8 +1,28 @@
...
+  syntax Bool ::= p1 ( Int ) [function, total, no-evaluators, klabel("p1"), symbol]
+                | p2 ( Int ) [function, total, no-evaluators, klabel("p2"), symbol]
+                | p3 ( Int ) [function, total, klabel("p3"), symbol]
+
+  // an indirect simplifiation loop
+  rule [p1-if-p2]: p1(X) => true requires p2(X) [simplification]
+  rule [p2-if-p1]: p2(X) => true requires p1(X) [simplification]
+  // simplification of terms involving p1(_someInt) or p2(_someInt) will
+  // enter a loop of p1, p2, p1, p2... until depth 100 is reached, and then
+  // bail out in booster. Simplification in kore-rpc _loops to infinity_.
```
jberthold, line 18: This can be tested with `booster-dev` - it recurses 100 times and then terminates with `TooManyIterations`. There is no loop being detected because it does not actually ever change the term. To see the loop detection kick in, we could create a test that uses `Int` arithmetic (and an llvm dy.lib) or some other simplifications  that happen before we enter the loop.

goodlyrottenapple (approved):

PR #262, opened 2023-08-09
Extend matching to support concrete maps

geo2a:
This PR:
* allows matching empty `KMap`s
* allows matching identical concrete `KMap`s


ehildenb:
Tests! Always add tests when adding to unification/matching! It's way too easy to make mistakes here.

goodlyrottenapple (approved):

PR #263, opened 2023-08-11
Fix a couple memory leaks

dwightguth:
Blocked on #264 

We were never taking any rewrite steps, so the garbage collector was never getting called. We also forgot to call kllvm_init, which meant that some of the memory that ought to have been allocated on the garbage collected heap in the llvm backend was instead being allocated in the wrong place.

Finally, we forgot to call `free` on the `char*` returned from `kore_simplify`.


library/Booster/LLVM/Internal.hs
```diff
@@ -342,9 +347,11 @@ mkAPI dlib = flip runReaderT dlib $ do
                                         }
                                 len <- fromIntegral <$> peek lenPtr
                                 cstr <- peek strPtr
-                                BS.packCStringLen (cstr, len)
+                                result <- BS.packCStringLen (cstr, len)
+                                Foreign.free cstr
```
goodlyrottenapple, line 351: Thanks for the spot here. I think I assumed that `packCStringLen` just wraps the pointer, but upon checking it copies the memory. I wonder if it would be worth passing this pointer tot he parse function and freeing after to avoid this copy? Not sure how big the returned data can be.... perhaps something to investigate later

goodlyrottenapple (approved):
LGTM, provided the tests pass with latest K

library/Booster/LLVM/Internal.hs
dwightguth: It might be worth it if we see performance issues related to this in a profile, but the advantage of this approach is that the native memory that is allocated is all isolated to this particular component, and all memory that gets used moving forward is just regular haskell heap memory. So I don't think we should prematurely optimize here.

goodlyrottenapple (approved):

PR #264, opened 2023-08-14
update k to v6.0.61

goodlyrottenapple:


dwightguth (changes requested):
I don't really feel qualified to review this. I don't understand nix or the haskell code well enough. The only comment I have is that changes to binary files in git make the repository size go up rapidly. It looks like you're trying to regenerate the decision trees each time you run the test, so maybe we should do that instead of committing each version of the binary file? You're going to very rapidly make the git repository huge if you do it this way, which has consequences for the build times of everything downstream.

goodlyrottenapple:
Yes, the plan is to regenerate the decision trees. was going to do this in a separate PR but can do this tomorrow and add to the current one. Not sure who else to ask for a review this week. Maybe @ehildenb ? Everyone else I can think of is on PTO...

dwightguth:
yes, let's add it to the current PR because every PR that changes the tar.gz file is going to add another couple megabytes to the size of the repo.

Not sure what to do about the review situation. I just know the changes to the nix files are all essentially gibberish to me and you could have done something really weird and I would have no way of knowing, so I don't really want to approve it. Is there not anyone with nix experience who could review that code that is not on vacation? I could probably review the rest of the changes if you add a bit of documentation to each file explaining what you're doing and why.

flake.nix
```diff
@@ -2,9 +2,7 @@
   description = "hs-backend-booster";
 
   inputs = {
-    k-framework.url = "github:runtimeverification/k/v5.6.131";
```
goodlyrottenapple, line 5: The changes to the nix flake are all to do with removing a direct dependence on the k-framework inside the flake. This is due to the fact that depending on the k-framework introduces a transitive cyclic dependency, which nix is not happy with and it makes maintaining this flake difficult. Instead, in the CI run, we simply build and add k to the PATH temporarily when running the integration tests

.github/workflows/test.yml
```diff
@@ -210,16 +210,23 @@ jobs:
...
-          nix run .#hs-backend-booster:test:llvm-integration --extra-experimental-features 'nix-command flakes' --print-build-logs
           export JQ=$(nix-build '<nixpkgs>' -A jq --no-link)/bin/jq
-          for dir in $(cd test/rpc-integration && ls -d test-*); do
+          K_VERSION=$(cat deps/k_release)
+          PLUGIN_VERSION=$(cat deps/blockchain-k-plugin_release)
+          export PATH="$(nix build github:runtimeverification/k/v$K_VERSION#k.openssl.procps --no-link --json | $JQ -r '.[].outputs | to_entries[].value')/bin:$PATH"
+          nix run .#hs-backend-booster:test:llvm-integration --extra-experimental-features 'nix-command flakes' --print-build-logs
+          
+          # The runDirectoryTest.sh script expects the following env vars to be set
+          export PLUGIN_DIR=$(nix build github:runtimeverification/blockchain-k-plugin/$PLUGIN_VERSION --no-link --json | $JQ -r '.[].outputs | to_entries[].value')
```
goodlyrottenapple, line 220: `nix build <drv> --no-link --json | $JQ -r '.[].outputs | to_entries[].value'` builds and returns the nix store path to `<drv>`

ehildenb (approved):

ehildenb:
The changes seem reasonable to me overall. I'm not an expert on the Nix, but a read of it doesn't show anything super surprising or anything. @dwightguth do you want to dismiss your review so we can merge?

dwightguth (approved):
This addressed my issue, so I'm approving. If Everett is happy with the rest of the code, go ahead and merge.

PR #269, opened 2023-08-16
Allow re-adding an already imported module

### LINKED ISSUE #267: add-module should allow the same module multiple times
This is the same issue as runtimeverification/haskell-backend#3640, however, it needs to be implemented in both backends.

goodlyrottenapple:
Fixes  #267

This change should allow pyk to launch a single server instance and run multiple proofs on the same instance, which involves each proof instance sending (potentially) the same add-module command to the server multiple times.


goodlyrottenapple:
already completed via #270

PR #272, opened 2023-08-21
Return `vacuous` when all applicable rules rewrite to bottom

### LINKED ISSUE #268: Return `vacuous` when state goes to `#bottom`.
This was implemented in the Haskell backend at https://github.com/runtimeverification/haskell-backend/pull/3637.  Opened the issue to keep track of the follow-up update.

goodlyrottenapple:
Fixes #268

This PR introduces a new `RewriteRuleAppT` monad which is analogous to the previously used `MaybeT`, but instead of returning only one kind of failure, namely `Nothing`, we can instead distinguish a failure with `NotApplicable` or `Trivial`. When we collect the results from attempting rule application, we return a `RewriteTrivial` if all the applied rules that we attempted resulted in the ensures clause of the rule evaluating to bottom.


.github/workflows/test.yml
```diff
@@ -218,14 +218,19 @@ jobs:
...
           cd test/rpc-integration
           for dir in $(ls -d test-*); do
               name=${dir##test-}
               echo "Running $name..."
-              ./runDirectoryTest.sh test-$name --time
+              if [ "$name" = "a-to-f" ]; then
+                  SERVER=BOOSTER_DEV ./runDirectoryTest.sh test-$name --time
+              else
+                  SERVER=KORE_RPC_BOOSTER ./runDirectoryTest.sh test-$name --time
+              fi
```
goodlyrottenapple, line 233: I wanted to make sure we run the `a-to-f` tests with `booster-dev`, to ensure we are catching the vacuous branch in booster and not in the old backbend. (this happened during testing when the code in the booster wasn't yet working)

library/Booster/Pattern/Rewrite.hs
```diff
@@ -128,17 +131,101 @@ rewriteStep cutLabels terminalLabels pat = do
...
+                -- all rules which did apply had an ensures condition which evaluated to false
+                | all (== Trivial) rxs ->
+                    -- if, all the other groups only generate a not applicable or trivial rewrites,
+                    -- then we return a `RewriteTrivial`. To ensure this, we need to catch the
+                    -- `NoApplicableRules` exception which is thrown when we have exhausted all the rules.
+                    -- When this error is thrown, we know that all the following rule groups had only
+                    -- non-applicable rules.
+                    catchNoApplicableRulesWhenTrivial (processGroups rest) >>= \case
+                        RewriteTrivial{} -> pure $ RewriteTrivial pat
+                        other -> pure other
```
jberthold, line 155: I don't really like catching exceptions here... maybe  `processGroups` could signal the condition as a `RewriteResult` ? We cannot directly return `RewriteTrivial`, though, as we would need to remember there was already a trivial result. So maybe `NoApplicableRules` should be `RewriteResult` instead of `RewriteFailed`?

library/Booster/Pattern/Rewrite.hs
```diff
@@ -487,6 +581,7 @@ performRewrite doTracing def mLlvmLibrary mbMaxDepth cutLabels terminalLabels pa
                 (Just _, [(lbl, uId, n)]) -> RewriteFinished (Just lbl) uId n
```
jberthold, line 581: Two lines above this one, `(Nothing, _) -> RewriteStuck p` should probably become `-> RewriteTrivial p` ?

Similarly for the 3 cases `Terminal`, `Finished`, `Aborted` below, which currently use `RewriteStuck`, too?

library/Booster/Pattern/Rewrite.hs
```diff
@@ -128,17 +128,89 @@ rewriteStep cutLabels terminalLabels pat = do
...
+                        RewriteStuck{} -> pure $ RewriteTrivial pat
+                        other -> pure other
+                -- at this point, there were some Applied rules and potentially some Trivial ones.
+                -- here, we just return all the applied rules in a `RewriteBranch`
+                | otherwise ->
+                    pure $
+                        RewriteBranch pat $
+                            NE.fromList $
+                                map (\(r, p) -> (labelOf r, uniqueId r, p)) $
+                                    concatMap (\case Applied x -> [x]; _ -> []) rxs
```
jberthold, line 157: Can this lead to pathological results where we return `RewriteBranch` but only have a single next state, e.g. for `results == [Applied something, Trivial, Trivial]`?
Maybe better to partition the `results` beforehand ? 
```
    let labelOf = ...
        (trivials, applieds) = partition (== Trivial) results
    if null results 
        then processGroups rest 
        else case applieds of
                [] -> ... -- all results were trivial for this group, proceed
                [Applied (r, x)] -> ... -- single result
                otherwise -> pure $ RewriteBranch pat applied
```

goodlyrottenapple: Ah, good catch!

jberthold (approved):

PR #279, opened 2023-08-24
Simplify `Stuck`/`Aborted` states with Kore before falling back to Kore's `execute`

geo2a:
This pr:
* runs the state from `Aborted`/`Stuck` response received from Booster through Kore's `simplify` before falling back to calling Kore's `execute`. This is a small change in [69a1c5e](https://github.com/runtimeverification/hs-backend-booster/pull/279/commits/69a1c5e578f90f4a7b18eb0eabb7b51e1b123f51).
* Renames and reshuffles some of the auxiliary functions in `Proxy.hs`. This will be useful in more content-intensive follow-up PRs

Part of #275


tools/booster/Proxy.hs
```diff
@@ -165,15 +163,17 @@ respondEither mbStatsVar booster kore req = case req of
                 -- in the old one to work around any potential
                 -- unification bugs.
                 | boosterResult.reason `elem` [Aborted, Stuck] -> do
```
goodlyrottenapple, line 165: what about the branching case?

goodlyrottenapple: do we want to address that in a separate PR?

geo2a: Yes, this will be addressed in a follow-up PR, along with re-running from pre-branch state after simplification.

tools/booster/Proxy.hs
```diff
@@ -219,6 +219,61 @@ respondEither mbStatsVar booster kore req = case req of
...
+                            , maxDepth = Just $ Depth 0
+                            }
+                Log.logInfoNS "proxy" "Making 0-step execute request to convert back to a term/constraints form"
+                result <- booster $ Execute request
+                case result of
+                    Right (Execute ExecuteResult{state = finalState}) -> pure finalState
+                    _other -> pure s
+            _other ->
+                -- TODO log error
+                pure s -- if we hit an error here, return the original
```
jberthold, line 247: Maybe add a log line here? (with level >= `warn`)

tools/booster/Proxy.hs
```diff
@@ -219,6 +219,61 @@ respondEither mbStatsVar booster kore req = case req of
...
+      where
+        toSimplifyRequest :: ExecuteState -> SimplifyRequest
+        toSimplifyRequest ExecuteState{term = t, substitution, predicate} =
+            let subAndPred = catMaybes [KoreJson.term <$> substitution, KoreJson.term <$> predicate]
+                termSort = KoreJson.SortApp (KoreJson.Id "SortGeneratedTopCell") []
+                term = t{KoreJson.term = foldr (KoreJson.KJAnd termSort) t.term subAndPred}
```
jberthold, line 253: This is basically `Booster.JsonRpc.execStateToKore`.
What occurred to me while reading that function is that we should try to use the sort of `t` instead of `SortGeneratedTopCell`.  We have `Booster.Syntax.Json.sortOfJson` but that one cannot give a result for `app` nodes. In principle we should check that all sorts in the `subAndPred` list are the same and use that, unless empty.

tools/booster/Proxy.hs
```diff
@@ -219,6 +219,61 @@ respondEither mbStatsVar booster kore req = case req of
             -- can only be an error at this point
             res -> pure res
 
+    simplifyExecuteState :: Maybe Text -> ExecuteState -> m ExecuteState
+    simplifyExecuteState mbModule s = do
+        Log.logInfoNS "proxy" "Simplifying execution state"
+        simplResult <-
+            handleSimplify $ toSimplifyRequest s
```
jberthold, line 226: If we use `handleSimplify` here, the statistics will include all those simplification calls as if they were regular requests. However, this may be the best alternative, otherwise we would need to reflect this `simplify-and-continue` in the statistics code.

jberthold (approved):
A few nit-picks but LGTM overall.
Approving under the assumption that this was already tested successfully on some foundry proofs.

tools/booster/Proxy.hs
geo2a: Done.

PR #281, opened 2023-08-25
Do not call LLVM backend on no-evaluators symbols

### LINKED ISSUE #273: Safe-guard against undefined functions in LLVM simplification
Currently, when functions without evaluators are sent to the LLVM backend simplification routine, the entire RPC server crashes after the routine prints the offending function name.

To address this in the Haskell code:
              * the translation to LLVM data structures should check for the `no-evaluators` attribute and refuse to translate function names that have this attribute. This could be implemented by modifying the (computed) attribute that checks whether a term is fully concrete to return `False` for such function calls.

_Originally posted by @jberthold in https://github.com/runtimeverification/haskell-backend/issues/3644#issuecomment-1688045231_

goodlyrottenapple:
This fixes #273 by reading the `no-evaluators` attribute of symbols and pre-computing the `canBeConcretelyEvaluated` attribute of terms, based on whether any of it's symbols have the `no-evaluators` attribute.


test/rpc-integration/test-foundry-bug-report/response-007.json
```diff
@@ -7789,7 +7789,7 @@
...
\u0001` \u0001a\u0011ÀV[\u0001`\u0004\u0001PPPV[`\u0000Qa\u0012þ` \u0001a\u0011ÀV[\u0001PPV[`\u0000` \u0003\u0012\u0015a\u0013\u001aW`\u0000ý[Q\u0015\u0015\u0014a\u0011\u0005W`\u0000ý[`@R`\u0000a\u0013T`@\u0001`\nRi\u0008\u0008\u0011^\u001c\u0019XÝ\u0019Y`²\u001b` \u0001R`@\u0001V[P` \u0001RPPV[`@R`\u0000a\u0013T`@\u0001`\nRi\u0008\u0008\u0008\u0008\u0010XÝ\u001dX[`²\u001b` \u0001R`@\u0001V[`@R`\u0000a\u0013¶`@\u0001`\nRi\u0008\u0008\u0011^\u001c\u0019XÝ\u0019Y`²\u001b` \u0001R`@\u0001V[\u0003` \u0001Ra\u0012@
a\u0011ðV[`@R`\u0000a\u0013¶`@\u0001`\nRi\u0008\u0008\u0008\u0008\u0010XÝ\u001dX[`²\u001b` \u0001R`@\u0001V[cNH{q`à\u001b`\u0000R`2`\u0004R`$`\u0000ý[`\u0000`\u0001\u0001a\u0014(WcNH{q`à\u001b`\u0000R`\u0011`\u0004R`$`\u0000ý[P`\u0001\u0001Vþ\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000q\tpÏ©\u001aboóhö[\u001dÑ-¢dipfsX\"\u0012 âÅêb¸jß©©.\u000cnã(F£²K¹\u0014\u0001=-Äo!ÉüdsolcC\u0000\u0008\r\u00003"
```
goodlyrottenapple, line 7792: these are just permutations inside a Map/list of constraits

library/Booster/LLVM.hs
```diff
@@ -29,7 +29,7 @@ simplifyTerm api def trm sort = unsafePerformIO $ Internal.runLLVM api $ do
     trmPtr <- Trace.timeIO "LLVM.simplifyTerm.marshallTerm" $ Internal.marshallTerm trm
     sortPtr <- Trace.timeIO "LLVM.simplifyTerm.marshallSort" $ Internal.marshallSort sort
     binary <- liftIO $ kore.simplify trmPtr sortPtr
-    liftIO $ kore.collect
+    liftIO kore.collect
```
jberthold, line 32: (off-topic) Interesting that this got accepted by `hlint`  - probably because the `kore.collect` desugars to a class function call (from `HasField`) before `hlint` gets to see it?

library/Booster/Pattern/Base.hs
```diff
@@ -113,6 +115,7 @@ data TermAttributes = TermAttributes
     -- variables, recursive through AndTerm
     , hash :: !Int
     , isConstructorLike :: !Bool
+    , canBeConcretelyEvaluated :: Bool
```
jberthold, line 118: (nit) why "concretely"? I think `canBeEvaluated` would be OK and shorter. ("evaluated" implying that we use LLVM and that it is concrete). But not too important

library/Booster/Pattern/Binary.hs
```diff
@@ -204,7 +204,7 @@ lookupKoreDefinitionSymbol name = DecodeM $ do
                     []
                     []
                     (SortApp "UNKNOWN" [])
-                    (SymbolAttributes PartialFunction IsNotIdem IsNotAssoc IsNotMacroOrAlias Nothing)
+                    (SymbolAttributes PartialFunction IsNotIdem IsNotAssoc IsNotMacroOrAlias Nothing (Flag False))
```
jberthold, line 207: Maybe define aliases `CannotBeEvaluated` and `CanBeEvaluated`?

library/Booster/Definition/Attributes/Base.hs
```diff
@@ -180,6 +180,7 @@ data SymbolAttributes = SymbolAttributes
     , isAssoc :: Flag "isAssoc"
     , isMacroOrAlias :: Flag "isMacroOrAlias"
     , isKMapSymbol :: Maybe KMapDefinition
+    , hasConcreteEvaluators :: Flag "hasConcreteEvaluators"
```
jberthold, line 183:
```suggestion
    , hasEvaluators :: Flag "hasEvaluators"
    , isKMapSymbol :: Maybe KMapDefinition
```

unit-tests/Test/Booster/Pattern/Util.hs
```diff
@@ -47,7 +47,7 @@ app :: Symbol -> [Term] -> Term
 app s = SymbolApplication s []
 
 asConstructor :: SymbolAttributes
-asConstructor = SymbolAttributes Constructor IsNotIdem IsNotAssoc IsNotMacroOrAlias Nothing
+asConstructor = SymbolAttributes Constructor IsNotIdem IsNotAssoc IsNotMacroOrAlias Nothing (Flag False)
```
jberthold, line 50: (probably not crucial in tests without LLVM backend but) a constructor won't ever have `hasEvaluators = Flag False` (`True` unless `no-evaluators` is present)

library/Booster/LLVM.hs
goodlyrottenapple: yeah i was surprised too

geo2a: For the record, `hlint` 3.6 flags it. Perhaps we should consider upgrading from 3.4?

geo2a (approved):
Approving, under the assumption that Jost's comments are addressed. Thanks @goodlyrottenapple!

library/Booster/Pattern/Base.hs
goodlyrottenapple: will change

PR #284, opened 2023-08-29
Add `kore-rpc-dev` --- `kore-rpc` with `kore-rpc-booster`-like interface

geo2a:


jberthold (approved):
LGTM.
Dependencies could be cleaned up but this is fine for the purpose

jberthold:
Should we run a few tests with the binary, too?

geo2a:
@jberthold I'm testing `kore-rpc-dev` on `test-a-to-f` as suggested, but unfortunately the workflow fails on Linux. I do not really understand why though, the same script works fine locally. I'll investigate further tomorrow.

jberthold:
> @jberthold I'm testing `kore-rpc-dev` on `test-a-to-f` as suggested, but unfortunately the workflow fails on Linux. I do not really understand why though, the same script works fine locally. I'll investigate further tomorrow.

https://github.com/runtimeverification/hs-backend-booster/actions/runs/6013238437/job/16310293317?pr=284#step:8:173 In the last run, the server did not start up cleanly, it probably failed to verify the definition for some reason(?).  Unfortunately we don't get the failure reason in the output, it is only in the tarball (maybe we can also do something about that in `kore-rpc` code?).

jberthold:
One thing to check maybe (for additional data): What does the test do if we don't set a `SERVER` variable at all, i.e., if we use `kore-rpc-booster`? Then the definition would be loaded by both booster and kore-rpc code, and probably also fail to load in the latter.

PR #287, opened 2023-08-30
Re-execute with Kore when Booster returns `Branching`

geo2a:
This PR re-runs in Kore, with depth 1, the execution request that Booster returns `Branching` for, similar to what we do for `Aborted` and `Stuck` results.

Note that getting `Branching` from Booster is likely impossible in KEVM, since Booster will return `Aborted` when it is unsure about rule unification. Once we have recursive simplification in Booster, we may actually get `Branching` though.

This PR adds a synthetic `test-diamond` integration tests that causes `booster-dev` to Branch, which `kore-rpc-booster` and `kore-rpc-dev` reach the terminal state. I've modified `runDirectoryTest.sh` to enable dedicated response files for `booster-dev` and `kore-rpc-dev`. `kore-rpc-booster` response files retain the normal `.json` extension.


test/rpc-integration/runDirectoryTest.sh
```diff
@@ -83,8 +83,16 @@ if [ -d $dir ]; then
...
         fi
         # call rpc-client
-        echo "$client $mode $test $params --expect $dir/response-${testname}.json $*"
-        $client $mode $test $params --expect $dir/response-${testname}.json $*
+        if [ $(basename $server) == "booster-dev" ]; then
+            echo "$client $mode $test $params --expect $dir/response-${testname}.booster-dev $*"
+            $client $mode $test $params --expect $dir/response-${testname}.booster-dev $*
+        elif [ $(basename $server) == "kore-rpc-dev" ]; then
+            echo "$client $mode $test $params --expect $dir/response-${testname}.kore-rpc-dev $*"
+            $client $mode $test $params --expect $dir/response-${testname}.kore-rpc-dev $*
```
goodlyrottenapple, line 91: should we have this branch? Given the contract of the booster-proxy is that it returns the same response as kore-rpc, the only two responses should be for booster-dev or booster-prox/kore-rpc

geo2a: Yes, you're right. However, the responses of  `kore-rpc-booster` and `kore-rpc-dev` do differ:
Screenshot_2023-09-04_12-05-42.

Do you think if would be easy to harmonize the responses of `kore-rpc-booster` and `kore-rpc-dev`? As I understand, `substitution` is more specialized than `predicate`, i.e. we could alter `kore-rpc`/`kore-rpc-dev` to always return substitution as predicate, just like `kore-rpc-booster` does? Or is it more complicated?

goodlyrottenapple: given we have kore-rpc-dev which wraps kore-rpc into a version more compatible with the booster, and we only use this for testing, we should just modify the response there

test/rpc-integration/test-diamond/response-infeasible-branching.kore-rpc-dev
```diff
@@ -0,0 +1 @@
...:"terminal-rule","depth":3,"rule":"TEST.BD","state":{"term":{"format":"KORE","version":1,"term":{"tag":"App","name":"Lbl'-LT-'generatedTop'-GT-'","sorts":[],"args":[{"tag":"App","name":"Lbl'-LT-'k'-GT-'","sorts":[],"args":[{"tag":"App","name":"kseq","sorts":[],"args":[{"tag":"App","name":"inj","sorts":[{"tag":"SortApp","name":"SortState","args":[]},{"tag":"SortApp","name":"SortKItem","args":[]}],"args":[{"tag":"DV","sort":{"tag":"SortApp","name":"SortState","args":[]},"value":"d"}]},{"tag":"App","name":"dotk","sorts":[],"args":[]}]}]},{"tag":"App","name":"Lbl'-LT-'int'-GT-'","sorts":[],"args":[{"tag":"DV","sort":{"tag":"SortApp","name":"SortInt","args":[]},"value":"42"}]},{"tag":"App","name":"Lbl'-LT-'generatedCounter'-GT-'","sorts":[],"args":[{"tag":"DV","sort":{"tag":"SortApp","name":"SortInt","args":[]},"value":"0"}]}]}},"substitution":{"format":"KORE","version":1,"term":{"tag":"Equals","argSort":{"tag":"SortApp","name":"SortInt","args":[]},"sort":{"tag":"SortApp","name":"SortGeneratedTopCell","args":[]},"first":{"tag":"EVar","name":"X","sort":{"tag":"SortApp","name":"SortInt","args":[]}},"second":{"tag":"DV","sort":{"tag":"SortApp","name":"SortInt","args":[]},"value":"42"}}}}}}
```
jberthold, line 1: ... interesting - `X` in the configuration is substituted by value `42` but then we return a substitution with `X` :question:

test/rpc-integration/runDirectoryTest.sh
jberthold: I guess having this case is a temporary necessity until we actually get response parity.
We could try to identify "substituting" predicates in booster and return them in the `substitution` field. Then again, what are they really for? Why do we not just substitute term and predicate? (and it seems we do, so the substitution is kind of "informational" for the caller to see what happened to the `X` - see other comment)

.github/workflows/test.yml
```diff
@@ -227,9 +227,10 @@ jobs:
           for dir in $(ls -d test-*); do
               name=${dir##test-}
               echo "Running $name..."
-              if [ "$name" = "a-to-f" ]; then
+              if [ "$name" = "diamond" ]; then
                   SERVER=$BOOSTER_DEV ./runDirectoryTest.sh test-$name --time
                   SERVER=$KORE_RPC_DEV ./runDirectoryTest.sh test-$name --time
```
jberthold, line 232: We should probably keep running the `a-to-f` test with both `booster-dev` and `kore-rpc-dev` ?

test/rpc-integration/runDirectoryTest.sh
geo2a: To clarify: in general, we should `#And` the substitution to the `predicate` field and set the `substitution` field to `Nothing`?

jberthold (approved):

.github/workflows/test.yml
geo2a: Done.

test/rpc-integration/test-diamond/response-infeasible-branching.kore-rpc-dev
geo2a: Investigation in https://github.com/runtimeverification/hs-backend-booster/issues/290

PR #292, opened 2023-09-05
286 internalise K Sets

### LINKED ISSUE #286: Internalize K sets

jberthold:
Fixes #286 

An internal K Set is a structure `KSet [Term] (Maybe Term)` with a list of concrete set elements (sorted by hash but _not deduplicated_ ) and an optional rest term (of compatible sort, which cannot be `KSet` by construction).
If a set is constructed using set concatenation of several sets containing such opaque rest terms, the concatenation is pushed inwards (i.e., all concrete elements are gathered in the sorted list, and the optional rest is the set concatenation of the opaque rest terms).

No attempt is made to match or unify these internal structures (yet). 

Set concatenation is partial as concatenating two sets with common elements results in an undefined term. No attempt is made to check for common/duplicate elements when internalising a set; the internalisation retains those common elements as duplicates in the list of concrete elements.


test/internalisation/test-totalSupply-definition.kore.report
```diff
@@ -1980,7 +1980,7 @@ Symbols: 1720
   - inj
   - kseq
 Axioms: 352
-Axioms preserving definedness: 295
+Axioms preserving definedness: 292
```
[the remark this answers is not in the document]
jberthold: This was caused by an imprecise analysis in `filterTermSymbols` (the concat symbol was checked in cases of singleton sets where it would not be required).

test/internalisation/test-totalSupply-definition.kore.report
```diff
@@ -1980,7 +1980,7 @@ Symbols: 1720
   - inj
   - kseq
 Axioms: 352
-Axioms preserving definedness: 295
+Axioms preserving definedness: 296
```
jberthold, line 1983: This change is caused by eliminating a redundant `.Set _Set_ OTHERSET` in a rule (by way of internalisation).

The rule has a `<activeAccounts> SetItem(ACCT) => .Set ... </activeAccounts>` (this field does not exist any more in the current evm-semantics config, btw).
In kore, the `...` becomes a variable `VAR`, and the  `.Set ...` on the RHS becomes `_Set_(.Set, VAR)`. Internalisation recognises the redundancy and removes the `_Set_` call; the rule is now marked as  preserving definedness.

library/Booster/Definition/Attributes/Reader.hs
```diff
@@ -211,6 +211,20 @@ instance HasAttributes ParsedSort where
...
+                                ( KCollectionSymbolNames
+                                    { unitSymbolName
+                                    , elementSymbolName
+                                    , concatSymbolName
+                                    }
+                                , KSetTag
+                                )
+                        }
             (Just _, Just _, Just _, Just _) ->
                 -- ignore any other hooked sorts like lists/sets
```
geo2a, line 229: Is this comment outdated now?
```suggestion
```

geo2a (approved):

library/Booster/Pattern/Base.hs
```diff
@@ -396,6 +404,64 @@ externaliseKList def heads optRest
...
+                    | Nothing <- rest1 ->
+                        KSet def elems1 (Just other2)
+                    | Just r1 <- rest1 ->
+                        KSet def elems1 (Just $ SymbolApplication concatSym [] [r1, other2])
+                (other1, KSet def2 elems2 rest2)
+                    | def2 /= def -> error $ "Inconsistent set definition " <> show (def2, def)
+                    | Nothing <- rest2 ->
+                        KSet def elems2 (Just other1)
+                    | Just r2 <- rest2 ->
+                        KSet def elems2 (Just $ SymbolApplication concatSym [] [other1, r2])
```
goodlyrottenapple, line 442: Won't this cause looping?? In the other cases for map/list we have something like:
```suggestion
                        KSet def elems2 (Just $ Term attribs $ SymbolApplicationF concatSym [] [other1, r2])
```

goodlyrottenapple: or perhaps we want to use `stripCollectionMetadata` as in the case below

library/Booster/Pattern/Util.hs
```diff
@@ -199,18 +202,29 @@ filterTermSymbols check = cata $ \case
         let concatSym = concatSymbol $ KListMeta def
             elemSym = klistElementSymbol def
             unitSym = unitSymbol $ KListMeta def
-         in if null heads
-                then [unitSym | check unitSym]
-                else filter check [concatSym, elemSym] <> concat heads
+         in case heads of
```
goodlyrottenapple, line 205: should this not apply to map as well? i.e. `KMap _ [x] Nothing` only needs to `check x`

library/Booster/Pattern/Util.hs
```diff
@@ -199,18 +202,29 @@ filterTermSymbols check = cata $ \case
...
+                [] -> [unitSym | check unitSym]
+                [x] -> [elemSym | check elemSym] <> x
+                more -> filter check [concatSym, elemSym] <> concat more
```
goodlyrottenapple, line 208: we don't really need to check any of the symbols in list since we know they are always total.

library/Booster/Pattern/Util.hs
```diff
@@ -199,18 +202,29 @@ filterTermSymbols check = cata $ \case
...
+        let concatSym = concatSymbol $ KSetMeta def
+            unitSym = unitSymbol $ KSetMeta def
+            elemSym = klistElementSymbol def
+         in case elements of
+                [] -> fromMaybe [unitSym | check unitSym] rest
+                [single] ->
+                    single
+                        <> maybe [elemSym | check elemSym] (filter check [concatSym, elemSym] <>) rest
+                more ->
+                    filter check [concatSym, elemSym] <> fromMaybe [] rest <> concat more
```
goodlyrottenapple, line 227:
```suggestion
                [] -> fromMaybe [] rest
                [single] ->
                    single
                        <> maybe [] (concatSym:) rest
                more ->
                    [concatSym] <> fromMaybe [] rest <> concat more
```

library/Booster/Pattern/Util.hs
jberthold: Good point. True for the currently single use of `filterTermSymbols`. In general, `check`  can be anything, though, so I would prefer to leave as-is.

library/Booster/Pattern/Util.hs
jberthold: I was not sure but thinking about it, yes, it is the same logic as in sets. Will adapt this case as well.

library/Booster/Pattern/Base.hs
jberthold: I don' t think it can loop because we are applying concatSym to new, smaller arguments (`r2` resp. `r1` above are strictly subsets of the originally-internalised set). In the case below we might basically reproduce the original term so we need to protect against loops there.
Of course I can protect against the possibility in these two cases as well but I don't think they can loop.

library/Booster/Pattern/Util.hs
jberthold: This is again because only `concatSym` is not total and the other two are?
I would like to keep the current code so it remains general. Otherwise we could specialise it to returning the partial symbols (i.e., hard-wire the `check` argument)  and rename `s/filterTermSymbols/partialTermSymbols`

PR #294, opened 2023-09-06
Split `substitution` from `predicate` when externalising `Pattern`

### LINKED ISSUE #290: Clarify the meaning and use of `substitution` in the execute response
The `substitution` filed of `ExecuteResponse` is returned by Kore when we fall-back to it in Booster. The substitution returned by Kore seems to be "informational", i.e. the variables will already have been substituted in the term. We, however, conjunct the substitution to the term when returning from Booster. This is likely unsound, since we add constraints on variables that are not present in the configuration anymore.

geo2a:
This PR changes how `kore-rpc-booster` externalises patterns when returning `ExecuteResult` to be in line with `kore-rpc`. Specifically: equalities between a variable on the left-hand-side and a term on the right-hand-side are returned in the `substitution` field, rather than as part of `predicate`.


library/Booster/Syntax/Json/Externalise.hs
```diff
@@ -10,33 +10,48 @@ module Booster.Syntax.Json.Externalise (
...
+                else Just . multiAnd sort . map (externalisePredicate sort) $ predicateItems
+     in (externaliseTerm term, predicate, substitution)
   where
     multiAnd :: Syntax.Sort -> [Syntax.KorePattern] -> Syntax.KorePattern
     multiAnd _ [] = error "multiAnd: empty"
     multiAnd sort ps = foldl1 (Syntax.KJAnd sort) ps
 
+    isSubstitutionItem :: Internal.Predicate -> Bool
+    isSubstitutionItem (Internal.EqualsTerm lhs _rhs) = isVariable lhs
+    isSubstitutionItem (Internal.AndPredicate lhs rhs) = isSubstitutionItem lhs && isSubstitutionItem rhs
```
geo2a, line 48: @jberthold @goodlyrottenapple I've added a recursive case here, but I'm not sure it is actually needed or even wanted. What do you think? `Pattern` holds a list of `Predicate` constraints. Could any of those `Predicate`'s be nested `AndPredicate`?

goodlyrottenapple: That shouldn't happen but probably best keep it in. I'm wondering whether we need to handle cases like `X ==Int ... #Equals true`?

goodlyrottenapple (approved):

library/Booster/Syntax/Json/Externalise.hs
geo2a: I though about it, and I don't know actually. I think these are actually eliminated when we run the the Booster result through the Kore simplifier. I will need to test it.

jberthold: Agree with Sam that we should keep this (we cannot guarantee this invariant for all future times).

jberthold (approved):
We can merge this as a first step in the right direction (response parity).
However, I think we could take a different approach, namely to identify substitution-like constraints in the internal format and then _apply_ the substitution as well as return it.

That would probably be part of the final simplification before returning the `execute` result. An even more invasive change would be to carry a substitution within the `Pattern` type which is _accumulated_ as we rewrite the term (accommodates the substitution that is returned by `kore-rpc` when we fall back).
We do not have to extend the internal `Pattern` type with a substitution field if we only handle substitutions at the end.

geo2a:
Thanks @jberthold! I've opened an issue to keep track of your suggestion: #297.

Issue #254, opened 2023-08-04
Matching on K/KItem sort injections


goodlyrottenapple:
@dwightguth has pointed out a case when a function should not be applied, but the booster does apply the owise case. Namely:

given

```
syntax Bool ::= isInt(K) [function]
rule isInt(I:Int) => true
rule isInt(_) => false [owise]
```

and a configuration with `isInt(X:KItem)`, the booster rewrites this term into `false`. Here is the simp trace for the application of this function:

```
isInt(VarX:SortKItem{})
Symbols differ:
kseq(Eq#VarInt:SortInt{}, dotk())
VarX:SortKItem{} @(hs-backend-booster-0.0.0-inplace:Booster.Pattern.ApplyEquations library/Booster/Pattern/ApplyEquations.hs:488:5)
[Simplify] Simplifying term
isInt(VarX:SortKItem{})
to
"false"
```

The issue above is the special treatment of injections into K, namely if we look at the kore definition of `isInt` and its internalisation, we get these two equations:

```
isInt(kseq{}(inj{SortInt{}, SortKItem{}}(Var'Unds'Gen1:SortInt{}),dotk{}())) = \dv{SortBool{}}("true")
isInt(VarK:SortK{})                                                          = \dv{SortBool{}}("false")
```

This is problematic because when applying the equations to the term `isInt(inj{SortKItem{}, SortK{})(VarX:SortKItem{}))`, the match between `kseq{}(..)` and `VarX:SortKItem{}` obviously fails... Not sure if this is an issue with how pyk generates kore for the K term `isInt(X:KItem)` (possibly related to https://github.com/runtimeverification/pyk/issues/565, i have not tested with a new enough version of pyk which contains the fix for this) or if the booster is supposed to handle the `kseq(..., dotk{})` patterns as a special case when matching...

dwightguth:
kseq and dotk are just regular constructors. What is supposed to happen, if I understand correctly, is that the first rule doesn't apply because pattern matching fails, but the second rule also doesn't apply because unification of the pattern with the first rule succeeds nontrivially.

goodlyrottenapple:
I would assume that the booster should determine that the match between the first pattern `isInt(I:Int)` and `isInt(X:KItem)` is indeterminate because `KItem` is a super-sort of `Int` and could therefore apply. Then, because the first equation is indeterminate, no other equations should be attempted to be matched and we should abort.

Is this intuition correct?

dwightguth:
[...]

PR #300, opened 2023-09-07
Check subsorting when internalising a kore term from a request

goodlyrottenapple:
This PR addresses part of #254, namely, the backend will now reject terms like `isInt{}(inj{SortKItem{}(), SortK{}()}(...))` because `SortKItem` is not declared a subsort of `SortK`.


library/Booster/Syntax/Json/Internalise.hs
```diff
@@ -130,10 +160,14 @@ internaliseTermRaw qq allowAlias sortVars definition@KoreDefinition{sorts, symbo
                 from' <- lookupInternalSort' from
                 to' <- lookupInternalSort' to
                 arg' <- recursion arg
+                unless (not (coerce checkSubsorts) || from' `isSubsort` to') $
```
goodlyrottenapple, line 163: We have to disable this check when internalising the kore definition, otherwise we would have to re-structure the module internalisation into two passes, one where we only internalise sorts and sub-sorts and then the second pass internalises all other axioms. Depends on how much we trust the front-end to emit correct injections :P

Given that we now have three flags, I decided to newtype them to avoid stupid mix-ups when passing the parameters around.

jberthold (approved):
:+1:

geo2a (approved):

PR #302, opened 2023-09-08
Address incompleteness of retuned rewriting traces

geo2a:
Needs https://github.com/runtimeverification/hs-backend-booster/pull/307

These PR fixes several places when we were not returning (or not fully returning) rewriting traces in `kore-rpc-booster`. I've covered the addressed cases with integration tests that demonstrate the differences (if any) between `booster-dev` (only Booster), `kore-rpc-dev` (only Kore) and `kore-rpc-booster` (both Booster and Kore). The added integration tests are:
* test/rpc-integration/test-a-to-f/state-terminal-rule-with-logging.json --- test which rewrite rules are applied in a **straight-line execution**
  - responses must be the same (two rewrites in Booster) for all three servers, see test/rpc-integration/test-a-to-f/response-terminal-rule-with-logging.json
* test/rpc-integration/test-diamond/params-mutual-constraints-terminal.json --- test which rewrite rules are applied **in presence of dependent constraints**
  - `booster-dev` fails immediately
  - `kore-rpc-dev` rewrites with two rules (pruning the infeasible branch)
  - `kore-rpc-booster` fails in Booster and then rewrites with Kore, like `kore-rpc-dev`
  - there's a variant of this test that does not specify terminal rules, turning the results into `stuck`. The rewrites are the same.

NOTE: integration tests do not cover simplification tracing.

Integration tests need to be finalized before we could merge.


library/Booster/JsonRpc.hs
```diff
@@ -512,10 +500,10 @@ mkLogRewriteTrace
                                 , origin = Booster
                                 }
             RewriteSimplified equationTraces Nothing
-                | fromMaybe False logSuccessfulSimplifications || fromMaybe False logFailedSimplifications ->
-                    mapM (mkLogEquationTrace equationLogOpts) equationTraces
+                | logSuccessfulSimplifications || logFailedSimplifications -> mapM (mkLogEquationTrace equationLogOpts) equationTraces
+                | otherwise -> Just []
```
geo2a, line 504: This case was missing and the call of `mkLogRewriteTrace` on a `RewriteSimplified` returned `Nothing` if `logSuccessfulSimplifications || logFailedSimplifications` was `False`. In other words, some rewriting traces were filtered-out if simplification logs were not requested.

test/rpc-integration/test-diamond/response-mutual-constraints-stuck.json
```diff
@@ -0,0 +1,101 @@
...
+                                    },
+                                    "value": "0"
+                                }
+                            ]
+                        }
+                    ]
+                }
+            }
+        },
+        "logs": []
```
geo2a, line 99: We don't seem to be getting logs if we revert to Kore and it gets stuck. This is likely a bug somewhere in the glue code, because we do get a rewrite success log in test/rpc-integration/test-diamond/response-mutual-constraints-terminal.json which rewrites using the same rule.

tools/booster/Proxy.hs
```diff
@@ -197,6 +217,7 @@ respondEither mbStatsVar booster kore req = case req of
                                     ( currentDepth + boosterResult.depth + koreResult.depth
                                     , time + bTime + kTime
                                     , koreTime + kTime
+                                    , maybe rpcLogs (rpcLogs ++) (liftA2 (++) boosterResult.logs koreResult.logs)
```
geo2a, line 220: We must pass the traces from previous iterations and return them in the final `ExecuteResponse`

tools/booster/Proxy.hs
```diff
@@ -216,74 +245,93 @@ respondEither mbStatsVar booster kore req = case req of
...
+            simplResult <-
+                handleSimplify $ toSimplifyRequest s
+            case simplResult of
+                -- This request should not fail, as the only possible
+                -- failure mode would be malformed or invalid kore
+                Right (Simplify simplified) -> do
+                    -- to convert back to a term/constraints form,
+                    -- we run a trivial execute request (in booster)
+                    -- We cannot call the booster internaliser without access to the server state
+                    -- Again this should not fail.
```
jberthold, line 281: That's the convincing comment I mentioned yesterday... :)  fails if ` KJBottom _ <- simplified.state`.
(my PR will conflict with this one but yours should go first)

tools/booster/Proxy.hs
```diff
@@ -216,74 +245,93 @@ respondEither mbStatsVar booster kore req = case req of
...
+                    let request =
+                            (emptyExecuteRequest simplified.state)
+                                { _module = mbModule
+                                , maxDepth = Just $ Depth 0
+                                }
+                    Log.logInfoNS "proxy" "Making 0-step execute request to convert back to a term/constraints form"
+                    result <- booster $ Execute request
+                    case result of
+                        Right (Execute ExecuteResult{state = finalState, logs = finalLogs}) ->
+                            pure (finalState, finalLogs)
```
jberthold, line 291: We would want to retain `simplified.logs` from the simplification response here. 
The `finalLogs` will always be empty because the 0-step request does not do any rewrites and all has been simplified before (or _should have been_...)

tools/booster/Proxy.hs
```diff
@@ -216,74 +245,93 @@ respondEither mbStatsVar booster kore req = case req of
...
+                    , _module = Nothing
+                    , cutPointRules = Nothing
+                    , terminalRules = Nothing
+                    , movingAverageStepTimeout = Nothing
+                    , stepTimeout = Nothing
+                    , logSuccessfulSimplifications
+                    , logFailedSimplifications
+                    , logSuccessfulRewrites
+                    , logFailedRewrites
+                    }
```
jberthold, line 321: Not necessary IMHO

tools/booster/Proxy.hs
```diff
@@ -216,74 +245,93 @@ respondEither mbStatsVar booster kore req = case req of
...
 
-    postExecSimplify :: Maybe Text -> API 'Res -> m (API 'Res)
-    postExecSimplify mbModule = \case
+    postExecSimplify ::
+        (Maybe Bool, Maybe Bool, Maybe Bool, Maybe Bool) -> Maybe Text -> API 'Res -> m (API 'Res)
```
jberthold, line 324: Maybe we should define `dataLogSettings = LogSettings { fieldnames... }` inside this file?

tools/booster/Proxy.hs
```diff
@@ -295,8 +343,13 @@ respondEither mbStatsVar booster kore req = case req of
...
                             res{reason = Stuck, nextStates = Nothing}
-                    _otherReason ->
-                        res{state = simplifiedState, nextStates}
+                    _otherReason
+                        | null filteredNexts ->
+                            res{state = simplifiedState, nextStates = Nothing}
+                        | otherwise ->
+                            res{state = simplifiedState, nextStates = Just filteredNexts}
+            -- NOTE: we deliberately ignore simplification traces produced by simplifyExecuteState.
+            --       append them to result.logs if needed in the future.
```
jberthold, line 352: :raised_eyebrow: Why that? I think the simplification logs must be appended (just like we collect them inside  `RewriteSimplified` during/after  rewrites, see [related code](https://github.com/runtimeverification/hs-backend-booster/blob/main/library/Booster/Pattern/Rewrite.hs#L580) )

tools/booster/Proxy.hs
geo2a: Good catch! I will add simplification logs and ignore the 0-step execution logs

geo2a: On second though, I think we should retain the 0-step logs too, just in case...

tools/booster/Proxy.hs
geo2a: Yeah, good point. Will add them.

tools/booster/Proxy.hs
```diff
@@ -295,8 +343,13 @@ respondEither mbStatsVar booster kore req = case req of
                         | null filteredNexts ->
                             -- HACK. Would want to return the prior state
                             res{reason = Stuck, nextStates = Nothing}
-                    _otherReason ->
-                        res{state = simplifiedState, nextStates}
+                    _otherReason
+                        | null filteredNexts ->
+                            res{state = simplifiedState, nextStates = Nothing}
+                        | otherwise ->
+                            res{state = simplifiedState, nextStates = Just filteredNexts}
```
geo2a, line 353: It turns out, we were not returning the simplified states, but rather the ones before simplification. Now here comes the fun part:
* returning `filteredNexts` in `Branching`, as originally intended, breaks an integration tests, specifically https://github.com/runtimeverification/hs-backend-booster/blob/main/test/rpc-integration/test-no-evaluator/
* The test brakes because `#Not (#Exists X. X #Equals 12)` is simplified into  `#Not (X #Equals 12)` (provided I understand the kore json corectly).

I will isolate this change into a separate PR.

tools/booster/Proxy.hs
geo2a: I've decided to not do it in this PR, as it already has too much changes for my taste... should probably do it when we touch this code the next time, but I think it's not urgent.

tools/booster/Proxy.hs
```diff
@@ -300,11 +351,30 @@ respondEither mbStatsVar booster kore req = case req of
...
+    xs -> Just xs
+  where
+    combineLogsImpl :: [RPCLog.LogEntry] -> [Maybe [RPCLog.LogEntry]] -> [RPCLog.LogEntry]
+    combineLogsImpl acc = \case
+        [] -> acc
+        (Nothing : rest) -> combineLogsImpl acc rest
+        (Just logs : rest) -> combineLogsImpl (acc ++ logs) rest
+
+appendLogs :: ExecuteResult -> Maybe [RPCLog.LogEntry] -> ExecuteResult
+appendLogs ExecuteResult{..} newLogs = ExecuteResult{logs = combineLogs [logs, newLogs], ..}
```
jberthold, line 380: I am a bit scared when using `RecordWildCards`, it silently introduces variable bindings that may cause subtle bugs.
In this case, a record pun would work fine: `appendLogs res = res { logs =- combineLogs [rec.logs, newLogs] }`.
But we can change this "next time we touch the code" (i.e., today :) )

jberthold (approved):

tools/booster/Proxy.hs
geo2a: Yes, I agree, let's get rid of `RecordWildCards` if possible. I could not find a better way to make the record update unambiguous, but I lack experience.

PR #309, opened 2023-09-13
303 return vacuous when fallback simplifies to bottom

### LINKED ISSUE #303: Stuck instead of vacuous result after fall-back to kore-rpc
Originally [reported by @PetarMax on slack](https://runtimeverification.slack.com/archives/CC360GUTG/p1694302074030139) (where he also supplies a bug report): 

When running (part of) a proof with `kore-rpc-booster`, execution stops with `stuck` , with a state where two constraints are contradicting each other (literally `\and(P, \not(P)) `). 
When run with `kore-rpc`, the result is correctly determined to be `vacuous`.

One aspect of this is that the simplification before the `fall-back` hits an error in the 0-step execution because the simplification result is a literal `\bottom` (execution requires the input state to contain at least one _term_).
With a fix to catch this problematic state , `kore-rpc` fall-back returns `stuck` with this literal `\bottom` as the state, instead of `vacuous` with the (unsimplified) prior state. In order to return `vacuous` with a meaningful state (not the literal `\bottom`), the `Proxy.hs` will have to catch this particular case where simplification yields `\bottom` and return the prior state (adds complexity to the fall-back implementation).

jberthold:
This specifically handles cases where combinations of constraints make a configuration undefined (`\bottom`), as observed in #303 . The post-execution simplification should yield `vacuous` in such cases but was previously returning `stuck`.

Responses differ between `kore-rpc-dev` and `kore-rpc-booster` because `kore-rpc-dev` will detect the contradiction during the rewrite step (which includes simplification) and not perform the rewrite in such cases. Also, it returns `stuck` instead of `vacuous` if  the initial state is already `\bottom` (not to be addressed in this code). 

Also includes some refactoring for the previous code change in `Proxy.hs` (simpler handling of logs, dedicated type for log settings).

Fixes #303

### DIFF
--- a/test/rpc-integration/test-vacuous/README.md
+++ b/test/rpc-integration/test-vacuous/README.md
@@ -1,0 +1,53 @@
+# Tests for `kore-rpc-booster` returning vacuous results after simplification
+
+Since `kore-rpc-booster` does not consider combinations of constraints, it won't discover when a reached state (or the initial state) simplifies to `\bottom`. In such cases, the `execute` request should return the current state with `"reason": "vacuous"` in the result.
+
+The `diamond/test.k` semantics is used, which consists of simple state
+transition rules featuring additional constraints on a variable
+`X:Int` in the configuration.
+
+Rules `init` and `AC` introduce constraints on this variable:
+
+* Rule `init => a` ensures `X ==Int 42`
+* Rule `a => c` ensures `X =/=Int 42`
+
+1) _vacuous-at-branch_
+
+   _Input:_
+   `execute` request with initial state `<k>init</k><int>0</int>`
+
+   _Expected:_
+   - Rewrite with `init` rule creates a false constraint `0 ==Int 42`
+     because `X ==Int 0` initially.
+   - The rewrite proceeds to the branch (`a => b` or `a => c`)
+   - Both branches are discarded because of the contradiction, the
+     result is `vacuous` with state `a`.
+
+   With `kore-rpc-dev`, the term won't be rewritten because the contradiction
+   will be recognised during the rewrite (simplification after each step).
+
+1) _vacuous-var-at-branch_ Same as 1) but using a variable `N` for the
+   `int` field and a constraint `N ==Int 0`.
+
+1) _vacuous-without-rewrite_
+
+   _Input:_
+   - `execute` request with initial state  `<k>d</k><int>N</int> \and N
+     ==Int 1  \and N =/=Int  1` (A contradiction in the initial constraints).
+
+   _Expected:_
+   - The rewrite is stuck with `<k>d</k><int>N</int> \and...(contradiction)`
+   - The result is simplified and discovered to be `vacuous` (with state `d`).
+1) _vacuous-but-rewritten_
+
+   _Input:_
+   - `execute` request with initial state  `<k>b</k><int>N</int> \and N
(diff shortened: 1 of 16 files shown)


test/rpc-integration/test-vacuous/README.md
```diff
@@ -0,0 +1,53 @@
+# Tests for `kore-rpc-booster` returning vacuous results after simplification
```
geo2a, line 1: Great write-up!!

tools/booster/Proxy.hs
```diff
@@ -324,57 +336,69 @@ respondEither mbStatsVar booster kore req = case req of
...
-    [] -> Nothing
-    xs -> Just xs
-  where
-    combineLogsImpl :: [RPCLog.LogEntry] -> [Maybe [RPCLog.LogEntry]] -> [RPCLog.LogEntry]
-    combineLogsImpl acc = \case
-        [] -> acc
-        (Nothing : rest) -> combineLogsImpl acc rest
-        (Just logs : rest) -> combineLogsImpl (acc ++ logs) rest
+combineLogs logSources
+    | all isNothing logSources = Nothing
```
geo2a, line 394: 👍 great refactoring!

tools/booster/Proxy.hs
```diff
@@ -324,57 +336,69 @@ respondEither mbStatsVar booster kore req = case req of
...
+                            (mapM $ simplifyExecuteState logSettings mbModule)
+                            nextStates
+                    let (logsOnly, (filteredNexts, filteredNextLogs)) =
+                            second unzip $ partitionEithers simplifiedNexts
+                        newLogs = simplifiedStateLogs : logsOnly <> filteredNextLogs
+
+                    pure $ case reason of
+                        Branching
+                            | null filteredNexts ->
+                                res{reason = Stuck, nextStates = Nothing}
```
geo2a, line 367: I think we're losing logs again here? Since we return the result from every brunch of the case expression, we need to include logs here to.
```suggestion
                                res{reason = Stuck, nextStates = Nothing, logs = combineLogs $ res.logs : simplifiedStateLogs : logsOnly}
```

tools/booster/Proxy.hs
```diff
@@ -275,28 +280,35 @@ respondEither mbStatsVar booster kore req = case req of
...
+                        case result of
+                            Right (Execute ExecuteResult{state = finalState}) ->
+                                -- return state converted by Booster and logs from simplification.
+                                -- The 0-step execute result won't have logs (no logs are requested)
+                                pure $ Right (finalState, simplified.logs)
+                            other -> do
+                                Log.logWarnNS "proxy" $
+                                    "Error in pseudo-execute step after simplification: "
+                                        <> either (Text.pack . show) (Text.pack . show . encode) other
+                                pure $ Right (s, Nothing)
```
goodlyrottenapple, line 307: perhaps we should just thread the server state here and do the conversion only? might be worth if we need to do a bigger refactor of the proxy?

geo2a: I agree that this should be done at some point, but at a low priority.

jberthold: Agree. I would just hope that we don't have to do that simplification business at all... but having access to the server state would be cleaner.

geo2a (approved):

PR #311, opened 2023-09-13
Log term on failing rewrite rewrite

### LINKED ISSUE #295: Produce a structured log of pre- and  post-states when Booster aborts and falls back to Kore
Kontrol proofs often cause `kore-rpc-booster` to abort in Booster and fall-back to Kore several times during an `execute` request. More and more abort cases require deep analysis to determine what exactly caused the abort. @ehildenb suggest implementing a tool to extract focused K claims (which need only a few rewrite steps to be proved) from large Kontrol integrations with `kore-rpc-booster`. While the tool itself will be implemented at `kontrol`/`pyk` level, `kore-rpc-booster` needs to provide data for it.

## Data that we need to produce 

What we want to know is:
* the pre-state, i.e. the one lead to Booster aborting;
* the post-state, i.e. the one Kore gave us back after executing for one step.

## How to produce that data

We want the two states to be ready-for-consumption by `pyk` , so we need to externalize them to KoreJson. Modifying the response data to make them available along with rewriting and simplification logs seems too heavy-weight (these logs seem to not be used at that point anyway...) Therefore I suggest logging them at a custom `BoosterAbort` log-level into `stderr` or a provided file. This way we would be able to have the following flow:

```
$ kore-rpc-booster definition.kore --llvm-backend-library interpreter.so -l Abort > file.out &
$ rpc-client send request.json
$ # filter file.out for BoosterAbort JSON messages
$ kevm generate-claim file.out
```

geo2a:
This is a sister PR to https://github.com/runtimeverification/haskell-backend/pull/3657, together with which it closes #295

### DIFF
--- a/library/Booster/JsonRpc.hs
+++ b/library/Booster/JsonRpc.hs
@@ -468,10 +468,11 @@ mkLogRewriteTrace
                                         , substitution = Nothing
                                         , ruleId = maybe "UNKNOWN" getUniqueId uid
                                         }
+                                , originalTerm = Nothing
                                 , origin = Booster
                                 }
             RewriteBranchingStep _ _ -> Nothing -- we may or may not want to emit a trace here in the future
-            RewriteStepFailed reason
+            RewriteStepFailed originalPattern reason
                 | logFailedRewrites ->
                     Just $
                         singleton $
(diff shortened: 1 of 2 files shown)


geo2a:
We've decided to pursue the alternative approach, i.e. emitting the abort log entries separately.

PR #320, opened 2023-09-29
Update `deps/k_release`

geo2a:


gtrepta:
https://github.com/runtimeverification/hs-backend-booster/actions/runs/6380667892/job/17315497608?pr=320#step:8:709

The blockchain-k-plugin_release needs to be updated in tandem with the k_release here, to account for a changed symbol name in the llvm-backend that it uses.

goodlyrottenapple (approved):

PR #324, opened 2023-10-03
Update booster to use multi-ary \and/\or

dwightguth:
This PR (https://github.com/runtimeverification/haskell-backend/pull/3666) introduces changes to kore-rpc-types, in particular the structure of KJAnd and KJOr. Here we update the booster to make use of these new data types. Once again, the internal representation of And and Or inside the backend is unchanged.


geo2a (commented):
Looks good to me, thanks @dwightguth! Let's wait until @goodlyrottenapple takes a look and then merge.

geo2a (approved):

PR #333, opened 2023-10-06
Siplmify and substitute unification remainders when applying rules

geo2a:
This PR modifies the `performRewrite` function in the following way:
* if any of the rule applications failed with `RuleApplicationUnclear` (unification is unclear), we get the unification remainders, attempt to simplify and substitute them into the term, and then retry applying rules
* mild refactoring to make the state in `performRewrite` a little more manageable

The diagram preceeding `perfromRewrite` needs to be updated. @goodlyrottenapple could you have a look at the code and also let me know which tool you used to produce the diagram?


goodlyrottenapple:
I used https://asciiflow.com/

library/Booster/Pattern/Util.hs
```diff
@@ -96,6 +97,28 @@ substituteInTerm substitution = goSubst
...
+        KMap attrs keyVals rest ->
+            fromMaybe
+                (KMap attrs (bimap goSubst goSubst <$> keyVals) (goSubst <$> rest))
+                (Map.lookup t substitution)
+        KList def heads rest ->
+            fromMaybe
+                (KList def (map goSubst heads) (bimap goSubst (map goSubst) <$> rest))
+                (Map.lookup t substitution)
+        KSet def elements rest ->
+            fromMaybe (KSet def (map goSubst elements) (goSubst <$> rest)) (Map.lookup t substitution)
```
goodlyrottenapple, line 120: just a slight niggle, but could refactor as

```haskell
goSubst t = flip fromMaybe (Map.lookup t substitution) $ case t of
  Var{} -> t
  DomainValue{} -> t
  SymbolApplication sym sorts args -> SymbolApplication sym sorts $ map goSubst args
 ...
```

library/Booster/Pattern/Rewrite.hs
```diff
@@ -483,58 +487,61 @@ showPattern title pat = hang 4 $ vsep [title, pretty pat.term]
...
-|              |              |             |               v  v        | |                                       |                 |               |
-|              |              |             |                           | +-----------------+                     |                 v               |
-|              |              |             |        Return vacuous P   |                   |                     |                                 |
-|              |              |             |                           |                   |                     |         Return branching        |
+|              +----------------------------+             v              v                    v                 v          |       v             |  |
+|              |              |             |                                                                              |                     |  |
+|              v              v             v           P' == _+_    P' /= _+_           /\ PS == _+_      PS simplify to  |   PS simplify to  +-+  |
```
goodlyrottenapple, line 515: the software might have changed these : `P' == _+_`, they should be `P' == _|_`

goodlyrottenapple (approved):
LGTM! the diagram needs a tiny fix

library/Booster/Pattern/Rewrite.hs
geo2a: good catch, thanks!

PR #337, opened 2023-10-10
remove \left-assoc and \right-assoc of \and and \or

dwightguth:
Blocked on https://github.com/runtimeverification/k/pull/3679

This completes the changes to the booster with respect to multi-ary \and and \or. Once the \left-assoc and \right-assoc syntax are no longer being emitted by the frontend, we can remove them from the backends.


library/Booster/Syntax/ParsedKore/Parser.y
```diff
@@ -419,11 +403,6 @@ the result. Namely, \\and, \\or, \\implies, and \\iff. Designed to be passed to
...
 -}
 mkApply :: Token -> [Sort] -> KorePattern -> KorePattern -> KorePattern
-mkApply tok [sort] first second
-    | Token _ TokenAnd <- tok = KJAnd sort [first, second]
-    | Token _ TokenOr <- tok = KJOr sort [first, second]
-    | Token _ TokenImplies <- tok = KJImplies sort first second
-    | Token _ TokenIff <- tok = KJIff sort first second
 mkApply tok sorts@[_, _] first second
     | Token _ (TokenIdent _) <- tok = KJApp (mkId tok) sorts [first, second]
 mkApply other _ _ _ = error $ "mkApply: unsupported token " <> show other
```
jberthold, line 408: (nit-pick) This could now be inlined into its only use site at `mkAssoc` (and `mkAssoc` likewise I guess)

jberthold (approved):
Should be merged to finish this and unblock follow-up work.
At some point, we should also make the following changes in the `haskell -backend`
* update the kore syntax documentation (which currently features a `multi-or` within `\left-assoc` or `\right-assoc`)
* adapt the json schema file to reflect the field changes for `and` and `or`. (earlier PR)

PR #342, opened 2023-10-18
Slimmed down the nix distribution of booster

goodlyrottenapple:
Copy of https://github.com/runtimeverification/haskell-backend/pull/3677


cabal.project.freeze
```diff
@@ -5,23 +5,16 @@ constraints: any.Cabal ==3.6.3.0,
...
              any.QuickCheck ==2.14.3,
-             QuickCheck -old-random +templatehaskell,
              any.StateVar ==1.2.2,
              any.adjunctions ==4.4.2,
              any.aeson ==2.0.3.0,
-             aeson -cffi +ordered-keymap,
              any.aeson-pretty ==0.8.9,
-             aeson-pretty -lib-only,
              any.ansi-terminal ==0.11.4,
-             ansi-terminal -example +win32-2-13-1,
```
jberthold, line 16: The script `scripts/freeze-cabal-to-stack-resolver.sh` generates these flags so maybe we should not remove them?
Or is the `nix`-ified version not generating them?

goodlyrottenapple: My local version of cabal removed those and i dont really have a way to run the exact version as on CI. Since we don't use the freeze file for anything anyway (as far as i know) and solely depend on stack.yaml in CI, it doesn't matter much IMO and in the next haskell backend ci update job it will likely put those back anyway...

test/rpc-integration/test-logTiming/test.sh
```diff
@@ -28,5 +28,5 @@ done
 echo "Running a request which gets stuck, with logTiming enabled"
 ${client} \
     execute $dir/state-c.execute ${client_args} -O log-timing=true | \
-    $JQ 'del(.result.logs[].time)' | \
+    jq 'del(.result.logs[].time)' | \
```
goodlyrottenapple, line 31: since the script can now be invoked locally, I've removed `$JQ` and assume `jq` is on local path

cabal.project.freeze
jberthold: OK fair enough, if we run into problems with the script we'll notice it soon enough.
I thought we were using the freeze file in the nix build (which uses `cabal`)...

jberthold (approved):
Code LGTM, :crossed_fingers: hope CI agrees

cabal.project.freeze
goodlyrottenapple: We were before, but I have switched us back to using stack.yaml/stackage as source of truth for nix.

PR #344, opened 2023-10-20
Replace `llvm-backend-matching` with `llvm-kompile-matching`

goodlyrottenapple:
Now that we have `llvm-kompile-matching` (https://github.com/runtimeverification/k/pull/3659) in K, replacing `llvm-backend-matching`, we can replace the calls and remove `llvm-backend-matching` from the llvm repo


jberthold (approved):
:heavy_check_mark:

PR #349, opened 2023-10-26
tarball-compare tool + store constraints in a set

### LINKED ISSUE #325: Create `kore-rpc-differ`: a tool to check parity of `kore-rpc-*` server responses
This tool should call kore-rpc and booster on requests in a bug report and compare for response parity via internal data structures. this way we should avoid pointless encoding differences between the JSON responses and have a much more detailed diffing, e.g. do configs differ or only some side conditions?

jberthold:
## Store constraints  as `Set`s
The `constraints` field n `Pattern` as well as `requires` and  `ensures`  fields in `RewriteRule`   now hold  a `Set Predicate` to avoid ordering issues and duplication.

## `tarball-compare` tool

Closes #325 

Tool to compare two bug report tarballs (or directories containing an unpacked bug report) for correspondence. 
The following comparisons are performed:

* compare number of request/response files in the tarball
* compare definition.kore file in the tarball (top-level)
* Then, for each pair of requests or responses in the rpc_* directories:
  * compare file contents and file size of json files
  * compare the number of steps (depth) in responses to execute requests
  * internalise states in response data (based on the definition.kore file) and compare internal data (displaying diff if not equal)

The tool can work on:

* a single gzipped tarball containing other tarballs for many tests (when given a single argument)
* two tarballs or directories of unpacked bug report data, assumed to contain the same `rpc_*` directory names  (when given two arguments, each one either a tar file or a directory)


library/Booster/Pattern/Base.hs
```diff
@@ -723,7 +723,7 @@ makeBaseFunctor ''Predicate
 -- | A term (configuration) constrained by a number of predicates.
 data Pattern = Pattern
     { term :: Term
-    , constraints :: ![Predicate]
+    , constraints :: Set Predicate
```
geo2a, line 726: !Haskell noob question alert!
Why are we dropping the strictness annotation here?

geo2a (approved):
Looks good to me! I only have one question:

when we run `tarball-compare <single_archive.gz>`, what format the archive has to be it? I.e. how should it pair the bug reports to be compared?

library/Booster/Syntax/ParsedKore/Internalise.hs
```diff
@@ -901,11 +901,11 @@ expandAlias alias currentArgs
                     Def.Pattern
                         { term = Util.substituteInTerm substitution term
                         , constraints =
-                            Util.substituteInPredicate substitution <$> constraints
+                            Set.map (Util.substituteInPredicate substitution) constraints
                         }
 
 removeTops :: Def.Pattern -> Def.Pattern
-removeTops p = p{Def.constraints = filter (/= Def.Top) p.constraints}
+removeTops p = p{Def.constraints = Set.filter (/= Def.Top) p.constraints}
```
goodlyrottenapple, line 908: can call `delete` here

goodlyrottenapple (approved):

library/Booster/Pattern/Base.hs
jberthold: The `Set` type is strict by itself.
The bang `!` just puts demand on the entire data structure so it might actually make sense to keep it.
I will just put it back, just in case.

library/Booster/Syntax/ParsedKore/Internalise.hs
jberthold: Good point, there will only be one `Def.Top` in the set now. Will fix it

Issue #271, opened 2023-08-21
Investigate recursive constraint evaluation / possibly experiment with caching


jberthold:
Issues https://github.com/runtimeverification/hs-backend-booster/issues/255 and https://github.com/runtimeverification/hs-backend-booster/issues/243 are reopened because of reverting the recursive constraint evaluation.

The reason for the revert was a regression in test `ERC20Test.testAllowance` (see description in #265).

- [ ] Investigate what exactly the regression is (extended logging and analysis)
- [ ] Aim to undo the revert, applying fixes where necessary

jberthold:
I re-ran the proof (test `ERC20Test.testAllowance` in `foundry-demo`) and inspected the `Rewrite` and `Simplify` logs.
Nothing points to an obvious loop in the simplification but there are a number of recursive simplifications (up to 6 levels of recursion) that try to simplify very simple expressions (e.g. `SomeVariable <=Int 1234...567890`). Many equations are tried because the `<=Int` index is very common for simplification equations.
Maybe we should experiment with a cache to avoid re-trying these (often futile) equations over and over again.
For the time being, it does not seem like a good idea  to re-instate the recursive simplification code.

PR #352, opened 2023-10-31
271 bring back recursive simplification

jberthold:
- [X] recursive simplification of side conditions in rewrite rules and equations
- [X] re-vamp of simplification integration tests 
   - now executes with `booster-dev` only
   - added a test that would loop in `kore-rpc-booster` or `kore-rpc-dev` 
- [x] restrict recursion depth to a small value (not 100 as it is now)
- [x] add a cache to avoid a runtime blow-out as described in #271


jberthold:
Performance impact is unclear at the moment, pausing work on this.

PR #357, opened 2023-11-03
Add profiling scripts

goodlyrottenapple:
This is a first iteration of the profiling scripts based on the script from #355.

Simply run via `scripts/performance-tests-kevm.sh` or `scripts/performance-tests-kontrol.sh`

### DIFF
--- a/scripts/performance-tests-kevm.sh
+++ b/scripts/performance-tests-kevm.sh
@@ -43,8 +43,7 @@ git submodule update --init --recursive --depth 1 kevm-pyk/src/kevm_pyk/kproj/pl
 
 
 
-feature_shell "make poetry"
-feature_shell "poetry run -C kevm-pyk -- kevm-dist --verbose build plugin haskell --jobs 4"
+feature_shell "make poetry && poetry run -C kevm-pyk -- kevm-dist --verbose build plugin haskell --jobs 4"
 
 feature_shell "make test-prove-pyk PYTEST_PARALLEL=8 PYTEST_ARGS='--timeout 7200 -vv --use-booster' > $SCRIPT_DIR/kevm-$KEVM_VERSION-$FEATURE_BRANCH_NAME.log"
 
--- a/scripts/performance-tests-kevm.sh
+++ b/scripts/performance-tests-kevm.sh
@@ -7,6 +7,13 @@ SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
 
 MASTER_COMMIT="$(git rev-parse origin/main)"
 
+FEATURE_BRANCH_NAME="$(git rev-parse --abbrev-ref HEAD)"
+FEATURE_BRANCH_NAME="${FEATURE_BRANCH_NAME//\//-}"
+
+if [[ $FEATURE_BRANCH_NAME == "master" ]]; then
+  FEATURE_BRANCH_NAME="feature"
+fi
+
 # Create a temporary directory and store its name in a variable.
 TEMPD=$(mktemp -d)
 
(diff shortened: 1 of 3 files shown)


scripts/performance-tests-kevm.sh
```diff
@@ -0,0 +1,57 @@
...
+
+feature_shell "make test-prove-pyk PYTEST_PARALLEL=8 PYTEST_ARGS='--timeout 7200 -vv --use-booster' > $SCRIPT_DIR/kevm-$KEVM_VERSION-$FEATURE_BRANCH_NAME.log"
+
+if [ ! -e "$SCRIPT_DIR/kevm-$KEVM_VERSION-master-$MASTER_COMMIT.log" ]; then
+  master_shell "make test-prove-pyk PYTEST_PARALLEL=8 PYTEST_ARGS='--timeout 7200 -vv --use-booster' > $SCRIPT_DIR/kevm-$KEVM_VERSION-master-$MASTER_COMMIT.log"
+fi
+
+cd $SCRIPT_DIR
+grep ' call  ' kevm-$KEVM_VERSION-$FEATURE_BRANCH_NAME.log > kevm-$KEVM_VERSION-master-$MASTER_COMMIT.$FEATURE_BRANCH_NAME
+grep ' call  ' kevm-$KEVM_VERSION-master-$MASTER_COMMIT.log > kevm-$KEVM_VERSION-master-$MASTER_COMMIT.master
```
jberthold, line 56: Maybe we can use `sed` and cut out the `" call "` string to produce data which can also be fed to `gnuplot`?
`sed  -n -e 's/^\([0-9.]*\)s call  *.*\[\(..*\)\]$/\2 \1/p' <logfile>` produces lines with format `<test-name> <time-in-seconds>` (without the `src/.../test_bla` prefix).
We could also directly pass in the output of `grep` to the script, using `python3 compare.py <(grep ' call  ' ...) <(grep ' call  ' ...)`

scripts/performance-tests-kevm.sh
```diff
@@ -0,0 +1,57 @@
...
+cd $TEMPD
+git clone --depth 1 --branch $KEVM_VERSION https://github.com/runtimeverification/evm-semantics.git
+cd evm-semantics
+git submodule update --init --recursive --depth 1 kevm-pyk/src/kevm_pyk/kproj/plugin
+
+
+
+feature_shell "make poetry && poetry run -C kevm-pyk -- kevm-dist --verbose build plugin haskell --jobs 4"
+
+feature_shell "make test-prove-pyk PYTEST_PARALLEL=8 PYTEST_ARGS='--timeout 7200 -vv --use-booster' > $SCRIPT_DIR/kevm-$KEVM_VERSION-$FEATURE_BRANCH_NAME.log"
```
jberthold, line 48: With 8-fold parallelism, I am getting very high variability in the timings. When testing this PR's branch against master, we would expect there to be no significant differences but the measured differences range from 0.8 to 1.3 on my machine. (I come to think that this is because we recompile for every test, not because of the tests)
```suggestion
feature_shell "make test-prove-pyk PYTEST_PARALLEL=4 PYTEST_ARGS='--timeout 7200 -vv --use-booster' > $SCRIPT_DIR/kevm-$KEVM_VERSION-$FEATURE_BRANCH_NAME.log"
```

goodlyrottenapple: Interesting... I didn't observe this variation on my linux machine. Perhaps we can make the parallelism an optional argument with 4 as default

scripts/performance-tests-kontrol.sh
```diff
@@ -0,0 +1,89 @@
...
+KONTROL_VERSION='v0.1.49'
+
+SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
+
+MASTER_COMMIT="$(git rev-parse origin/main)"
+
+FEATURE_BRANCH_NAME=${FEATURE_BRANCH_NAME:-"$(git rev-parse --abbrev-ref HEAD)"}
+FEATURE_BRANCH_NAME="${FEATURE_BRANCH_NAME//\//-}"
+
+PYTEST_PARALLEL=${PYTEST_PARALLEL:-2}
```
jberthold, line 17: :+1: ... let's carry these changes over to the `kevm` test script, too.

jberthold (approved):
The `kontrol` script is in good shape now! Please carry over the changes to the `kevm` one, too. Pre-approving.

geo2a:
I've made the changes to the `kevm` script too, merging now.

Issue #321, opened 2023-09-29
Booster crash: custom hooked `Map` is stripped of the element sort after LLVM simplification


virgil-serbanuta:
I have a query that makes the booster backend stop with an error, while it works when querying the Haskell backend directly.

Bug report: [bug-report.zip](https://github.com/runtimeverification/hs-backend-booster/files/12765678/bug-report.zip)

Output:
```
...
[Info#proxy] Starting RPC server
[Info] Process request 1 execute
[Info#proxy] Starting execute request
[Rewrite] Stopped after 0 steps. @(hs-backend-booster-0.0.0-DpKnlYxxzPeBfB7OJwCE7f:Booster.Pattern.Rewrite library/Booster/Pattern/Rewrite.hs:559:18)
[Rewrite] No rules applicable for the pattern Term:
...
Conditions:
    \equalsTerm{}(_<=Int_("0", VarARG_0:SortInt{}), "true") @(hs-backend-booster-0.0.0-DpKnlYxxzPeBfB7OJwCE7f:Booster.Pattern.Rewrite library/Booster/Pattern/Rewrite.hs:559:18)
[Rewrite] Applied simplification @(hs-backend-booster-0.0.0-DpKnlYxxzPeBfB7OJwCE7f:Booster.Pattern.Rewrite library/Booster/Pattern/Rewrite.hs:559:18)
[Rewrite] Retrying with simplified pattern @(hs-backend-booster-0.0.0-DpKnlYxxzPeBfB7OJwCE7f:Booster.Pattern.Rewrite library/Booster/Pattern/Rewrite.hs:559:18)
[Rewrite] Stopped after 0 steps. @(hs-backend-booster-0.0.0-DpKnlYxxzPeBfB7OJwCE7f:Booster.Pattern.Rewrite library/Booster/Pattern/Rewrite.hs:559:18)
[Rewrite] No rules applicable for the pattern Term:
...
Conditions:
    \equalsTerm{}(_<=Int_("0", VarARG_0:SortInt{}), "true") @(hs-backend-booster-0.0.0-DpKnlYxxzPeBfB7OJwCE7f:Booster.Pattern.Rewrite library/Booster/Pattern/Rewrite.hs:559:18)
[Info#proxy] Booster Stuck at Depth {getNat = 0}
[Info#proxy] Simplifying booster state and falling back to Kore 
[Info#proxy] Simplifying execution state
[Info#booster] Simplifying a pattern
[...]

PR #363, opened 2023-11-09
Custom hooked `Map` externalisation

geo2a:
Work on  #321

Just adding integration tests for now.

~I'm struggling to reproduce #321 as the provided bug report unfortunately contains a pre-multy-or LLVM definition. I've requested access to the relevant code bases in order to produce obtain the up-to-date definition.~

This turns out to be an issue in the LLVM backend; hence, blocked on https://github.com/runtimeverification/llvm-backend/issues/886


test/llvm-integration/LLVM.hs
```diff
@@ -1301,6 +1380,42 @@ defSymbols =
                         }
                 }
             )
+        ,
+            ( "Lbl'Unds'Int2Int'Pipe'-'-GT-Unds'"
+            , Symbol
+                { name = "Lbl'Unds'Int2Int'Pipe'-'-GT-Unds'"
+                , sortVars = []
+                , argSorts = [SortApp "SortWrappedInt" [], SortApp "SortInt" []]
```
Baltoli, line 1388: @geo2a I don't follow why the second argument has sort `Int` here; in the corresponding definition the second argument has sort `WrappedInt`? I might be missing something though.

test/llvm-integration/definition/llvm.k
```diff
@@ -39,3 +40,30 @@ module LLVM
...
+  syntax MapInt2Int ::= MapInt2Int MapInt2Int
+         [ left, function, hook(MAP.concat), klabel(_MapInt2Int_),
+           symbol, assoc, comm, unit(.MapInt2Int), element(_Int2Int|->_),
+           index(0), format(%1%n%2)
+         ]
+  syntax MapInt2Int ::= ".MapInt2Int"
+         [ function, total, hook(MAP.unit),
+           klabel(.MapInt2Int), symbol, latex(\dotCt{MapInt2Int})
+         ]
+  syntax MapInt2Int ::= WrappedInt "Int2Int|->" WrappedInt
```
Baltoli, line 63: Here's the production I believe corresponds to the above comment @geo2a

test/llvm-integration/LLVM.hs
geo2a: You're right, and it indeed makes the test I constructed to pass. We have been able to pin-point the problem to the `MAP.update` hook in the LLVM backend. I'll update the integration tests and also open an issue in the LLVm backend with the somewhat minimized definition and input.

Baltoli: Remarkable timing; I had also just managed to build and get the same test to pass with the value sort as `WrappedInt`! Let's take a look at your new minimized example in our meeting this afternoon.

goodlyrottenapple:
closing in favour of #378

PR #377, opened 2023-11-20
`==K` simplification

### LINKED ISSUE #367: Simplification of `==K` in the booster
As @virgil-serbanuta pointed out in a recent discussion, the booster currently cannot reason over `==K` terms containing variables. I believe in the old backend, we use the following rule:

```
rule {K1 ==K K2 #Equals true} => {K1 #Equals K2} [simplification]
```
and then use reasoning over `#Equals`, which is built into the old backend. 

Ideally, the front-end should generate equations for `==K` for any given sort, corresponding to structural equality/structural equality+AC (I presume?? It's not even clear what `==K` is, given there are no equations for it as it is hooked in one way or the other in both back-ends) instead of relying on built-ins, as this makes things a lot more scrutable.

Obviously, a simpler alternative would be to just implement a custom simplifier in the booster and call it whenever we see `==K`, to deal with instances such as `CONSTR1( ... VarX ...) ==K CONSTR2(...)` which cannot be sent to the llvm backend but can easily be discharged as false, given `CONSTR1 /= CONSTR2`

goodlyrottenapple:
This PR adds custom simplification for `==K` predicates and should fix #367. The long term fix should probably be to add these rules/simplifications to the kore file (tracked by https://github.com/runtimeverification/k/issues/3810)


goodlyrottenapple:
```
| Test                                     | sam-eqauls-k-simplifier time | master time        | (sam-eqauls-k-simplifier/master) time
|------------------------------------------|------------------------------|--------------------|---------------------------------------
| mcd/vat-addui-fail-rough-spec.k          | 152.85                       | 274.22             | 0.5573991685507985
| erc20/hkg/transferFrom-failure-2-spec.k  | 130.82                       | 162.86             | 0.8032666093577304
| erc20/hkg/allowance-spec.k               | 72.41                        | 81.48              | 0.8886843397152675
| erc20/ds/approve-failure-spec.k          | 100.7                        | 110.62             | 0.9103236304465738
| erc20/ds/transferFrom-failure-2-c-spec.k | 166.7                        | 182.73             | 0.9122749411700323
| erc20/hkg/transfer-success-2-spec.k      | 73.01                        | 79.41              | 0.9194056164211057
| mcd/dstoken-approve-fail-rough-spec.k    | 177.07                       | 187.37             | 0.9450285531301702
| erc20/ds/transfer-success-2-spec.k       | 92.49                        | 95.98              | 0.9636382579704104
| mcd/vat-addui-pass-spec.k                | 266.67                       | 254.81             | 1.0465444841254268
| erc20/ds/transferFrom-failure-1-b-spec.k | 214.67                       | 203.1              | 1.0569670113244707
| examples/solidity-code-spec.md           | 256.27                       | 240.03             | 1.0676582093904927
| bihu/forwardToHotWallet-failure-4-spec.k | 143.76                       | 133.99             | 1.072915889245466
| erc20/ds/transferFrom-failure-2-a-spec.k | 108.76                       | 97.93              | 1.1105891963647503
| mcd/flipper-addu48u48-fail-rough-spec.k  | 100.48                       | 90.41              | 1.1113814843490766
| mcd/vat-slip-pass-rough-spec.k           | 324.82                       | 290.91             | 1.11656526073356
| erc20/hkg/transfer-success-1-spec.k      | 85.49                        | 75.5               | 1.1323178807947019
| erc20/hkg/totalSupply-spec.k             | 62.75                        | 52.75              | 1.1895734597156398
| mcd/vat-mului-pass-spec.k                | 287.48                       | 237.1              | 1.2124841838886546
| erc20/hkg/transferFrom-failure-1-spec.k  | 208.17                       | 158.02             | 1.3173648905201871
| mcd/dsvalue-peek-pass-rough-spec.k       | 137.82                       | 74.51              | 1.8496846060931416
| TOTAL                                    | 3163.19                      | 3083.7299999999996 | 1.0257674958572898
```

goodlyrottenapple:
```
| Test                                                                       | sam-eqauls-k-simplifi time | mast time          | (sam-eqauls-k-simplifi/mast) time
|----------------------------------------------------------------------------|----------------------------|--------------------|-----------------------------------
| LoopsTest.test_sum_10()                                                    | 90.72                      | 96.79              | 0.9372869098047318
| AssertTest.testFail_assert_false()                                         | 59.81                      | 63.11              | 0.9477103470131517
| StoreTest.testStoreLoadNonExistent()                                       | 79.33                      | 82.55              | 0.9609933373712901
| SymbolicStorageTest.testFail_SymbolicStorage(uint256)                      | 269.1                      | 278.87             | 0.9649657546527056
| src/tests/integration/test_foundry_prove.py::test_foundry_auto_abstraction | 60.87                      | 58.8               | 1.035204081632653
| AssumeTest.test_assume_true(uint256,uint256)                               | 73.88                      | 70.78              | 1.0437976829612885
| EmitContractTest.testExpectEmitCheckEmitter()                              | 118.98                     | 113.91             | 1.0445088227548065
| BytesTypeTest.test_bytes32(bytes32)                                        | 51.14                      | 48.92              | 1.0453802125919869
| ERC20.sol                                                                  | 54.06                      | 51.65              | 1.0466602129719265
| PlainPrankTest.test_startPrank_zeroAddress_true()                          | 183.52                     | 174.5              | 1.0516905444126075
| AccountParamsTest.testDealSymbolic(uint256)                                | 79.15                      | 74.97              | 1.0557556355875684
| MethodDisambiguateTest.test_method_call()                                  | 53.12                      | 50.3               | 1.0560636182902585
| EmitContractTest.testExpectEmitDoNotCheckData()                            | 122.06                     | 115.39             | 1.0578039691481065
| StoreTest.testGasStoreColdVM()                                             | 71.32                      | 67.06              | 1.0635252013122576
| AddrTest.test_addr_true()                                                  | 66.42                      | 62.27              | 1.0666452545366951
| ERC721.sol                                                                 | 55.89                      | 52.34              | 1.0678257546809322
| SymbolicStorageTest.testFail_SymbolicStorage1(uint256)                     | 321.24                     | 297.97             | 1.078095110245998
| TOTAL                                                                      | 1810.6100000000001         | 1760.1799999999996 | 1.0286504789282918
```

geo2a (approved):
Looks good! While reviewing, I've come up with a could of simple unit tests and allowed myself to push them.

unit-tests/Test/Booster/Pattern/ApplyEquations.hs
```diff
@@ -146,6 +147,39 @@ test_simplifyPattern =
...
+                        @?= Right
+                            (Predicate (EqualsK (KSeq someSort [trm| A:SomeSort{} |]) (KSeq someSort [trm| B:SomeSort{} |])))
+            , testCase "Different constructors, same variable" $
+                let subj =
+                        EqualsK (KSeq someSort [trm| con1{}(A:SomeSort{}) |]) (KSeq someSort [trm| con2{}(A:SomeSort{}) |])
+                 in simpl (Predicate subj) @?= Right (Predicate (DomainValue boolSort "false"))
+            , testCase "Different constructors, different variables" $
+                let subj =
+                        EqualsK (KSeq someSort [trm| con1{}(A:SomeSort{}) |]) (KSeq someSort [trm| con2{}(B:SomeSort{}) |])
+                 in simpl (Predicate subj) @?= Right (Predicate (DomainValue boolSort "false"))
```
goodlyrottenapple, line 173:
```suggestion
                 in simpl (Predicate subj) @?= Right (Predicate FalseBool)
```

library/Booster/Pattern/Base.hs
```diff
@@ -692,9 +692,18 @@ injectionSymbol =
         }
 
 -- convenience patterns
```
jberthold, line 694: (for discussion, not on this PR)
We have grown a large amount of these convenience patterns - and they are quite specific `K` idioms.
Maybe we should separate them into their own module (or modules? `Booster.Pattern.Bool` and `Booster.Pattern.Equalities`? But where would `KSeq` go? ) and document the assumptions we make in haddock comments.

library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -688,27 +688,64 @@ simplifyConstraints doTracing def mbApi cache ps =
...
+    foldAndBool (x : xs) = AndBool x $ foldAndBool xs
+
+    evalEqualsK l@(SymbolApplication sL _ argsL) r@(SymbolApplication sR _ argsR)
+        | isConstructorSymbol sL && isConstructorSymbol sR =
+            if sL == sR
+                then foldAndBool <$> zipWithM evalEqualsK argsL argsR
+                else pure FalseBool
+        | otherwise =
+            (if recurseIntoEvalBool then evalBool else pure) $
+                EqualsK (KSeq (sortOfTerm l) l) (KSeq (sortOfTerm r) r)
```
jberthold, line 737: I am not sure about this recursion, would it not go into a loop if `recurseIntoEvalBool` was true?
Going like this:
1. call to `evalBool` with argument `EqualsK (KSeq s1 (f x)) (KSeq s2 (g y))` with functions `f` and `g`
2. line 707 calls `evalEqualsK (f x) (g y)`
3. line 736 calls `evalBool (EqualsK (KSeq s1' (f x)) (KSeq s2' (g y)))` (arriving at step 1)

I think we should at least not recurse in this case here. The second case in line 751 appears similar, we are reproducing the original call that led us to this case.

library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -688,27 +688,64 @@ simplifyConstraints doTracing def mbApi cache ps =
...
+    evalEqualsK (SymbolApplication symbol _ _) _
+        | isConstructorSymbol symbol = pure FalseBool
+    evalEqualsK _ (SymbolApplication symbol _ _)
+        | isConstructorSymbol symbol = pure FalseBool
```
jberthold, line 741: Stumbling over these cases as well. We know the resp. other argument is not a `SymbolApp` but what about variables? A variable could stand for  _anything_ so we cannot be certain that the variable is _not_ instantiated to the exact same `SymbolApp` as the considered argument.
Not sure where this case would occur in practice but I don't think  we can return `FalseBool`.

library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -688,27 +688,64 @@ simplifyConstraints doTracing def mbApi cache ps =
...
+    t -> if recurseIntoEvalBool then evalBool t else pure t
   where
     evalBool :: MonadLoggerIO io => Term -> EquationT io Term
     evalBool t = do
         prior <- getState -- save prior state so we can revert
         result <- iterateEquations 100 TopDown PreferFunctions t
         EquationT $ lift $ lift $ put prior
         pure result
+
+    recursion = simplifyConstraint' recurseIntoEvalBool
```
jberthold, line 718: Used only once (`NotBool`), maybe inlne?

library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -688,27 +688,64 @@ simplifyConstraints doTracing def mbApi cache ps =
...
+
+    negateBool TrueBool = FalseBool
+    negateBool FalseBool = TrueBool
+    negateBool x = NotBool x
+
+    foldAndBool [] = TrueBool
+    foldAndBool [x] = x
+    foldAndBool (FalseBool : _) = FalseBool
+    foldAndBool (TrueBool : xs) = foldAndBool xs
+    foldAndBool (x : xs) = AndBool x $ foldAndBool xs
```
jberthold, line 728: This could probably go into said `Booster.Pattern.Bools` module if we end up splitting off those domain modules.

library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -688,27 +688,64 @@ simplifyConstraints doTracing def mbApi cache ps =
     runEquationT doTracing def mbApi cache $ mapM (simplifyConstraint' True) ps
```
jberthold, line 688:
```suggestion
    runEquationT doTracing def mbApi cache $ mapM (coerce . simplifyConstraint' True . coerce) ps
```
to fix the build...
(that line looks scary, though :fearful: )

library/Booster/Pattern/ApplyEquations.hs
goodlyrottenapple: Done

library/Booster/Pattern/ApplyEquations.hs
goodlyrottenapple: Thanks for the catch

library/Booster/Pattern/Base.hs
goodlyrottenapple: Created the suggested `Booster.Pattern.Bool` and moved the patterns plus functions like `splitBoolPredicates` to this module

PR #378, opened 2023-11-20
Fix incorrect KMap injections, received from the LLVM library

goodlyrottenapple:
This is a temporary fix for #321 until properly addressed upstream by https://github.com/runtimeverification/llvm-backend/issues/886 . The fix involves manually adjusting an injection to the correct sort, since we know the sort of the argument for a given symbol from the `KoreDefintion`, passed to the decoder.

### DIFF
--- a/test/llvm-integration/LLVM.hs
+++ b/test/llvm-integration/LLVM.hs
@@ -152,7 +152,6 @@ latin1Prop api = property $ do
                 | otherwise -> error $ "Unexpected sort " <> show s
             otherTerm -> error $ "Unexpected term " <> show otherTerm
 
-
 mapKItemInjProp :: Internal.API -> Property
 mapKItemInjProp api = property $ do
     let k = wrapIntTerm 1
@@ -179,12 +178,15 @@ mapKItemInjProp api = property $ do
 
     wrapIntTerm :: Int -> Term
     wrapIntTerm i =
-        SymbolApplication (defSymbols Map.! "inj") [SortApp "SortWrappedInt" [], SortApp "SortVal" []] [
-            SymbolApplication
+        SymbolApplication
+            (defSymbols Map.! "inj")
+            [SortApp "SortWrappedInt" [], SortApp "SortVal" []]
+            [ SymbolApplication
                 (defSymbols Map.! "LblwrapInt")
                 []
                 [intTerm i]
             ]
+
 ------------------------------------------------------------
 
 runKompile :: IO ()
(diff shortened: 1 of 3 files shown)


test/llvm-integration/LLVM.hs
```diff
@@ -90,6 +91,10 @@ llvmSpec =
                 it "should work with latin-1strings" $
                     hedgehog . propertyTest . latin1Prop
 
+            describe "Correct sort injections in non KItem maps" $
+                it "should work with latin-1strings" $
```
jberthold, line 95: Thinking about it again, ... We are running this 2000 times (`modifyMaxSuccesses (* 20)` above, default 100 AFAIK).
Maybe it would work to just repeat the `beforeAll loadAPI` here? And we can remove the `describe`
```suggestion
        beforeAll loadAPI $
            it "should correct sort injections in non KItem maps" $
```

test/llvm-integration/LLVM.hs
```diff
@@ -147,6 +152,41 @@ latin1Prop api = property $ do
                 | otherwise -> error $ "Unexpected sort " <> show s
             otherTerm -> error $ "Unexpected term " <> show otherTerm
 
+mapKItemInjProp :: Internal.API -> Property
+mapKItemInjProp api = property $ do
+    let k = wrapIntTerm 1
+    let v = wrapIntTerm 2
+    LLVM.simplifyTerm api testDef (update k v) (SortApp "SortMapValToVal" []) === singleton k v
```
jberthold, line 159: Maybe also add a test for the standard map? (although we should assume that one works :smirk: )

jberthold (approved):
Tests could be tweaked, but pre-approving

test/llvm-integration/LLVM.hs
```diff
@@ -147,6 +152,41 @@ latin1Prop api = property $ do
                 | otherwise -> error $ "Unexpected sort " <> show s
             otherTerm -> error $ "Unexpected term " <> show otherTerm
 
+mapKItemInjProp :: Internal.API -> Property
+mapKItemInjProp api = property $ do
+    let k = wrapIntTerm 1
+    let v = wrapIntTerm 2
```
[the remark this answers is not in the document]
goodlyrottenapple: I've gone with @jberthold's suggestion of calling only once with the fixed values, since we are not interested in the value serialisations

goodlyrottenapple:
I've confirmed that with the latest k version, removing our injection fix works for the unit test we originally added to test this behaviour. Do we want to revert the injection fix or leave it in case someone still uses an old version of the llvm-backend?

PR #382, opened 2023-11-21
Improve parsing of performance test logs

goodlyrottenapple:
This should allow us to store only the `.log` files and do the ` call    ` filtering inside the python script, which should help de-clutter the produced files as the number of log files grows quickly....

### DIFF
--- a/scripts/performance-tests-kevm.sh
+++ b/scripts/performance-tests-kevm.sh
@@ -10,6 +10,7 @@ KEVM_VERSION=${KEVM_VERSION:-'master'}
 SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
 
 MASTER_COMMIT="$(git rev-parse origin/main)"
+MASTER_COMMIT_SHORT="$(git rev-parse --short origin/main)"
 
 FEATURE_BRANCH_NAME=${FEATURE_BRANCH_NAME:-"$(git rev-parse --abbrev-ref HEAD)"}
 FEATURE_BRANCH_NAME="${FEATURE_BRANCH_NAME//\//-}"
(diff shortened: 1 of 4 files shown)


scripts/performance-tests-kevm.sh
```diff
@@ -54,12 +55,10 @@ feature_shell "make poetry && poetry run -C kevm-pyk -- kevm-dist --verbose buil
 feature_shell "make test-prove-pyk PYTEST_PARALLEL=$PYTEST_PARALLEL PYTEST_ARGS='--maxfail=0 --timeout 7200 -vv --use-booster' | tee $SCRIPT_DIR/kevm-$KEVM_VERSION-$FEATURE_BRANCH_NAME.log"
 killall kore-rpc-booster || echo "No zombie processes found"
 
-if [ ! -e "$SCRIPT_DIR/kevm-$KEVM_VERSION-master-$MASTER_COMMIT.log" ]; then
-  master_shell "make test-prove-pyk PYTEST_PARALLEL=$PYTEST_PARALLEL PYTEST_ARGS='--maxfail=0 --timeout 7200 -vv --use-booster' | tee $SCRIPT_DIR/kevm-$KEVM_VERSION-master-$MASTER_COMMIT.log"
+if [ ! -e "$SCRIPT_DIR/kevm-$KEVM_VERSION-master-$MASTER_COMMIT_SHORT.log" ]; then
+  master_shell "make test-prove-pyk PYTEST_PARALLEL=$PYTEST_PARALLEL PYTEST_ARGS='--maxfail=0 --timeout 7200 -vv --use-booster' | tee $SCRIPT_DIR/kevm-$KEVM_VERSION-master-$MASTER_COMMIT_SHORT.log"
```
jberthold, line 59: While we are at it, could we make a dedicated (`.gitignore`-d) directory for the log data?
We would replace `SCRIPT_DIR` here by `LOG_DIR` (defaults to `realpath ${SCRIPT_DIR}/../log-store` or similar), and `cd` into that dir below, calling `python3 ${SCRIPT_DIR}/compare.py`; likewise in the other script

scripts/performance-tests-kontrol.sh
```diff
@@ -10,6 +10,7 @@ KONTROL_VERSION=${KONTROL_VERSION:-'v0.1.49'}
 SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
```
jberthold, line 10: Also while we are here, `KONTROL_VERSION` is still hard-wired to 0.1.49 in this script and should be switched to `master` as in the `kevm` script (also changing how we build things).

jberthold (commented):
:+1: What is here LGTM. Suggested a few more tweaks

geo2a (approved):
I agree with Jost's suggestions, pre-approving!

scripts/performance-tests-kontrol.sh
```diff
@@ -62,15 +70,15 @@ master_shell() {
 
 feature_shell "poetry install && poetry run kevm-dist --verbose build plugin haskell foundry --jobs 4"
```
jberthold, line 71: Should the build targets get prefixed here, as in the `kevm` script?
```suggestion
feature_shell "poetry install && poetry run kevm-dist --verbose build evm-semantics.plugin evm-semantics.haskell foundry --jobs 4"
```

goodlyrottenapple: ah probably. im just testing them before merging so will correct

PR #400, opened 2023-12-01
Split hs-backend-booster package into dev-tools and main package

### LINKED ISSUE #374: Consistent names for binaries, remove dev tools from distro
Prompted by [this request from slack](https://runtimeverification.slack.com/archives/CC360GUTG/p1700167759618309):
* The binaries installed into `/usr/bin` by hs-backend-booster as part of the K distro should have consistent names prefixed with `kore-`. 
* Development tools should not be installed, they can probably go into a separate package that is not installed with K.

Currently all binaries in the `tools` directory are installed. Some of them (`*-dev`, `parsetest`) are needed for CI, some others are merely for manual developer use (or ephemeral). `rpc-client` should be called `kore-rpc-client`.

jberthold:
The main package is reduced to executables `kore-rpc-booster` and `kore-rpc-client` (renamed)
Other executables (`parsetest*`, `booster-dev`, `kore-rpc-dev`, `kore-json-differ`, `tarball-compare`, `eventlog-parser` ) are relocated to an extra package `hs-backend-booster-dev-tools`.

Once this is merged, the[ `stack install` command in the downstream repository `k` ](https://github.com/runtimeverification/k/blob/master/hs-backend-booster/pom.xml#L62-L66)can be changed from `stack install` to `stack install hs-backend-booster` to omit the extra binaries.

Fixes #374


geo2a (approved):

dev-tools/hs-backend-booster-dev-tools.cabal
```diff
@@ -0,0 +1,184 @@
+cabal-version: 1.12
```
goodlyrottenapple, line 1: the cabal file should not be checked in

goodlyrottenapple: if not, we should remove

```
-- This file has been generated from package.yaml by hpack version 0.35.2.
--
-- see: https://github.com/sol/hpack
```

jberthold: Good catch, thanks! Deleting this file (and adding it to `.gitignore`)

jberthold: Well well... that makes the nix build for the integration tests fail, though. Looking into it...

PR #405, opened 2023-12-04
Internalise #And under #Not

goodlyrottenapple:
This makes the predicate internalisation more permissive by allowing predicates of the form `#Not ( ... #And ... )` to be internalised as `notBool (... andBool ...)` note, any substitutions will become `==Int/==Bool/==K` predicates and the internaliser will fail if it encounters a `#Ceil`


geo2a (approved):
LGTM!

PR #411, opened 2023-12-06
Misc dev fixes

goodlyrottenapple:
This PR contains a number of small misc dev tools/fixes:

* Add `pretty`, a KORE JSON pretty printer util, useful when looking at bug reports without the original k definition (If we have the K definition, It's better to call pyk's pretty printer). The README has been updated with a sample use-case invocation.
* Fix performance scripts, `kevm-dist` -> `kdist`
* Modify the nix shell to automatically call `hpack` when the shell is re-loaded, useful to keep the local cabal files in sync with `package.yaml`
* Cleanup the README, removing references to stale nix stuff related to the dev shell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,37 +47,15 @@ There are several things you can do, to make the development via nix as seamless
 
 #### Nix shell
 
-To open the nix shell you will need nix version 2.4 or newer. Then use either `nix develop` (if you have flakes enabled) or use the old style `nix-shell` command.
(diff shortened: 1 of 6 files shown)


jberthold (approved):
LGTM :+1: 
We should probably also (at some point) implement a `Pretty` instance for `KoreJSON` to print directly without any internalisation/checks. That one should live in `kore-rpc-types`.

PR #414, opened 2023-12-07
Concurrent `kore-rpc-booster`

### LINKED ISSUE #383: Support multiple clients in parallel
Currently the `kore-rpc-booster` server is able to handle clients connecting to it and making requests from different processes, but at some point there is a bottleneck where only one connection at a time can make progress, preventing any speedup from using multiple clients to make independent requests in parallel to one server. From what I understood, there are at least 2 issues to be dealt with for this limitation to be overcome: opening multiple z3 processes, and the fact that the servers contain state from `add_module`, because of which @tothtamas28 had said we might consider different ways of handling the modules that get used in the other request types, e.g. sessions or cookies.

goodlyrottenapple:
Mirror PR to https://github.com/runtimeverification/haskell-backend/pull/3702
Fixes #383


goodlyrottenapple:
Closing in favour of #427

PR #416, opened 2023-12-11
384 409 rework predicate internalisation

### LINKED ISSUE #384: Apply more rigid criteria on valid substitutions
Current internalisation code identifies substitutions as every `\equals` term with only a variable on either side.
Not all such cases are actually useful or even valid "substitutions"
* `X:Bool \equals \notBool(X:Bool)`  (easy to solve by `==Bool`)
* `X:UserSort \equals f(Y:UserSort)`  (valid substitution)
* `X:UserSort \equals f(X:UserSort)`  (not a substitution)
* `X \equals f(Y) ;  Y \equals g(Z) ; Z \equals h(X)`  (substitutions need a direction, and here we probably need a topo-sort of all LHS and RHS terms in substitutions to discard this case as circular)

Related: #297

### LINKED ISSUE #409: Unable to internalise a nest of \not(\and(\ceil(_), \equals(_,_)))
When sending a term that contains a nest of `\not ( \and ( \ceil (something), \equalis (another, thing) ) )`  to the simplify end-point, the request fails with an internalisation error.

The [`foldPreds` code in `internalisePredicate`](https://github.com/runtimeverification/hs-backend-booster/blob/main/library/Booster/Syntax/Json/Internalise.hs#L362-L369) appears to be missing a case for this combination.

 [and-ceil-equals-request-crashing.json](https://github.com/runtimeverification/hs-backend-booster/files/13575367/and-ceil-equals-request-crashing.json) 

<details>
<summary>
The request can be sent to the simplify end-point using command

`kore-rpc-client simplify and-ceil-equals-request-crashing.json`

</summary>

```json
{
    "format": "KORE",
    "version": 1,
    "term": {
        "tag": "Not",
        "sort": {
            "tag": "SortApp",
            "name": "SortGeneratedTopCell",
            "args": []
        },
        "arg":
            {
                "tag": "And",
                "sort": {
                    "tag": "SortApp",
                    "name": "SortGeneratedTopCell",
                    "args": []
                },
                "patterns": [
                    {
                        "tag": "Ceil",
                        "argSort": {
                            "tag": "SortApp",
                            "name": "SortInt",
                            "args": []
                        },
                        "sort": {
                            "tag": "SortApp",
                            "name": "SortGeneratedTopCell",
                            "args": []
                        },
                        "arg": {
                            "tag": "DV",
                            "sort": {
                                "tag": "SortApp",
                                "name": "SortInt",
                                "args": []
                            },
                            "value": "0"
                        }
                    },
                    {
                        "tag": "Equals",
                        "argSort": {
                            "tag": "SortApp",
                            "name": "SortInt",
                            "args": []
                        },
                        "sort": {
                            "tag": "SortApp",
                            "name": "SortGeneratedTopCell",
                            "args": []
                        },
                        "first": {
                            "tag": "EVar",
                            "name": "VarX",
                            "sort": {
                                "tag": "SortApp",
                                "name": "SortInt",
                                "args": []
                            }
                        },
                        "second": {
                            "tag": "DV",
                            "sort": {
                                "tag": "SortApp",
                                "name": "SortInt",
                                "args": []
                            },
                            "value": "0"
                        }
                    }
                ]
            }
    }
}
```

</details>

<details> 
<summary> Response  to this request shows "Not supported" </summary>

```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": 2,
        "data": [
            {
                "context": [
                    {
                        "tag": "And",
                        "sort": {
                            "tag": "SortApp",
                            "name": "SortGeneratedTopCell",
                            "args": []
                        },
                        "patterns": [
                            {
                                "tag": "Ceil",
                                "argSort": {
                                    "tag": "SortApp",
                                    "name": "SortInt",
                                    "args": []
                                },
                                "sort": {
                                    "tag": "SortApp",
                                    "name": "SortGeneratedTopCell",
                                    "args": []
                                },
                                "arg": {
                                    "tag": "DV",
                                    "sort": {
                                        "tag": "SortApp",
                                        "name": "SortInt",
                                        "args": []
                                    },
                                    "value": "0"
                                }
                            },
                            {
                                "tag": "Equals",
                                "argSort": {
                                    "tag": "SortApp",
                                    "name": "SortInt",
                                    "args": []
                                },
                                "sort": {
                                    "tag": "SortApp",
                                    "name": "SortGeneratedTopCell",
                                    "args": []
                                },
                                "first": {
                                    "tag": "EVar",
                                    "name": "VarX",
                                    "sort": {
                                        "tag": "SortApp",
                                        "name": "SortInt",
                                        "args": []
                                    }
                                },
                                "second": {
                                    "tag": "DV",
                                    "sort": {
                                        "tag": "SortApp",
                                        "name": "SortInt",
                                        "args": []
                                    },
                                    "value": "0"
                                }
                            }
                        ]
                    }
                ],
                "error": "Pattern not supported"
            },
            {
                "context": [
                    {
                        "tag": "Not",
                        "sort": {
                            "tag": "SortApp",
                            "name": "SortGeneratedTopCell",
                            "args": []
                        },
                        "arg": {
                            "tag": "And",
                            "sort": {
                                "tag": "SortApp",
                                "name": "SortGeneratedTopCell",
                                "args": []
                            },
                            "patterns": [
                                {
                                    "tag": "Ceil",
                                    "argSort": {
                                        "tag": "SortApp",
                                        "name": "SortInt",
                                        "args": []
                                    },
                                    "sort": {
                                        "tag": "SortApp",
                                        "name": "SortGeneratedTopCell",
                                        "args": []
                                    },
                                    "arg": {
                                        "tag": "DV",
                                        "sort": {
                                            "tag": "SortApp",
                                            "name": "SortInt",
                                            "args": []
                                        },
                                        "value": "0"
                                    }
                                },
                                {
                                    "tag": "Equals",
                                    "argSort": {
                                        "tag": "SortApp",
                                        "name": "SortInt",
                                        "args": []
                                    },
                                    "sort": {
                                        "tag": "SortApp",
                                        "name": "SortGeneratedTopCell",
                                        "args": []
                                    },
                                    "first": {
                                        "tag": "EVar",
                                        "name": "VarX",
                                        "sort": {
                                            "tag": "SortApp",
                                            "name": "SortInt",
                                            "args": []
                                        }
                                    },
                                    "second": {
                                        "tag": "DV",
                                        "sort": {
                                            "tag": "SortApp",
                                            "name": "SortInt",
                                            "args": []
                                        },
                                        "value": "0"
                                    }
                                }
                            ]
                        }
                    }
                ],
                "error": "Pattern must contain at least one term"
            }
        ],
        "message": "Could not verify pattern"
    }
}
```

</details>

jberthold:
Predicates are now always internalised as a group, and classified into different kinds of predicates.

1. Input patterns are split up at top-level `\and`s.
2. predicate parts are classified into 
   * `BoolPred`icate (containing a term of `Bool` sort),
   * `CeilPred`icate, containing `\ceil(term)`
   * `SubstitutionPred`icate, containing an equation `\equals(var, term)` with a single variable on one side that is not present (i.e. free) on the other side
   * `UnsupportedPred`icate, which is ignored, left untouched, and re-added to the output
3. the set of `SubstitutionPred`icates is vetted to ensure it represents a valid non-circular and unambiguous substitution
   * Only one equation per variable is allowed (otherwise all those equations become `BoolPred`icates).
   * There must not be a circular dependency between any variables in the equations

it is an error to have any unsupported predicates in any part of a definition. 

Fixes #409 
Closes #384


library/Booster/JsonRpc.hs
```diff
@@ -99,7 +100,11 @@ respond stateVar =
                 Left patternError -> do
                     Log.logDebug $ "Error internalising cterm" <> Text.pack (show patternError)
                     pure $ Left $ RpcError.backendError RpcError.CouldNotVerifyPattern patternError
-                Right (pat, substitution) -> do
+                Right (pat, substitution, unsupported) -> do
+                    unless (null unsupported) $
+                        Log.logWarnNS
+                            "booster"
+                            "Execute: ignoring unsupported predicate parts"
```
goodlyrottenapple, line 107: any chance we could maybe make this a custom log level and print the offending term? I suspect this might be a source of "why isn't my state being pruned" issues, because we fail to internalise some constraint and thus don't send it to the solver...

library/Booster/Syntax/Json/Internalise.hs
```diff
@@ -269,41 +311,111 @@ internaliseTerm ::
...
+                pure [BoolPred $ Internal.Predicate $ Internal.NEqualsInt a b]
+            [BoolPred (Internal.Predicate (Internal.NEqualsInt a b))] ->
+                pure [BoolPred $ Internal.Predicate $ Internal.EqualsInt a b]
+            [BoolPred (Internal.Predicate (Internal.EqualsK a b))] ->
+                pure [BoolPred $ Internal.Predicate $ Internal.NEqualsK a b]
+            [BoolPred (Internal.Predicate (Internal.NEqualsK a b))] ->
+                pure [BoolPred $ Internal.Predicate $ Internal.EqualsK a b]
+            [BoolPred (Internal.Predicate p')] ->
+                pure [BoolPred $ Internal.Predicate $ Internal.NotBool p']
+            [SubstitutionPred k v] ->
```
goodlyrottenapple, line 403: what about cases where we have more than one `BoolPred` or `Substitution`? i.e. `#Not (... #And ...)` which we have seen previously

library/Booster/JsonRpc.hs
```diff
@@ -206,26 +215,38 @@ respond stateVar =
...
+                                                <> map (externaliseCeil predicateSort) (Set.toList ps.ceilPredicates)
+                                                <> map (uncurry $ externaliseSubstitution predicateSort) (Map.toList ps.substitution)
+                                                <> ps.unsupported
 
-                                pure $ Right (addHeader $ Syntax.KJAnd predicateSort result, traces)
-                            (Left something, _traces, _) ->
-                                pure . Left . RpcError.backendError RpcError.Aborted $ show something -- FIXME
+                                    pure $ Right (addHeader $ Syntax.KJAnd predicateSort result, traces)
+                                (Left something, _traces, _) ->
+                                    pure . Left . RpcError.backendError RpcError.Aborted $ show something -- FIXME
```
geo2a, line 249:
```suggestion
                                    pure . Left . RpcError.backendError RpcError.Aborted $ show something
```

library/Booster/Syntax/Json/Internalise.hs
```diff
@@ -269,41 +311,111 @@ internaliseTerm ::
...
-internalisePredicate ::
+{- | Internalises an And-ed set of predicates, classifying them into
+  BoolPred, CeilPred, SubstitutionPred, and UnsupportedPred.
+  The substitution (as a whole) is analysed after internalisation, to
+  ensure nothing is circular or ambiguous.
+
+  Throws a PatternError when any of the top-level patterns is a term
+  or an And with a term (must be analysed beforehand).
+-}
+internalisePredicates ::
```
geo2a, line 322: :exploding_head:

library/Booster/Syntax/Json/Internalise.hs
```diff
@@ -269,41 +311,111 @@ internaliseTerm ::
...
+        InternalisedPredicates
+            { boolPredicates = Set.fromList $ [p | BoolPred p <- internalised] <> moreEquations
+            , ceilPredicates = Set.fromList $ [p | CeilPred p <- internalised]
+            , substitution
+            , unsupported = [p | UnsupportedPred p <- internalised]
+            }
+  where
+    mkSubstitution ::
+        [InternalisedPredicate] -> (Map Internal.Variable Internal.Term, [Internal.Predicate])
+    mkSubstitution initialSubst =
```
geo2a, line 347: This code is quite cool! Perhaps worth factoring-out the `mkSubstitution` into a top-level function and unit-test it in isolation?

library/Booster/Syntax/Json/Internalise.hs
jberthold: All input like that currently go into the `_ -> notSupported` case  (creating an `UnsupportedPred`).
We could support `#Not ( ... #Or ... #Or ... )` for cases where all the `...` can be negated easily (the `Equals* <--> NEquals*`). As we discussed the other day this would mean to implement DeMorgan.
I don't think we can support `#Not ( ... #And ...)` unless the `...` are all of `Bool` sort (and likewise this would mean implementing DeMorgan with assumed `notBool` and `orBool` operations).

unit-tests/Test/Booster/Fixture.hs
```diff
@@ -85,15 +86,33 @@ app s = SymbolApplication s []
...
+                , symbolType = TotalFunction
+                , isIdem = IsNotIdem
+                , isAssoc = IsNotAssoc
+                , isMacroOrAlias = IsNotMacroOrAlias
+                , hasEvaluators = CanBeEvaluated
+                , smt = Just (SMTHook (Atom "="))
+                }
+        }
+kseq = [symb| symbol kseq{}(SortKItem{}, SortK{}) : SortK{} [constructor{}()] |]
+dotk = [symb| symbol dotk{}() : SortK{} [constructor{}()] |]
```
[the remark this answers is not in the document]
jberthold: Yeah, thought about that. But I wanted to explicitly have a different way to define those two so we will spot changes caused by how we internalise the two symbols.

library/Booster/Syntax/Json/Internalise.hs
jberthold: I moved this to top-level but the unit tests I wrote are still working on the `internalisePredicates` level.

library/Booster/Syntax/Json/Internalise.hs
jberthold: Added a description to `mkSubstitution` and a link to the gist

library/Booster/JsonRpc.hs
jberthold: I introduced a new `ErrorDetails` level to print more things whenever a `logWarn` or `logError` occurs.
For the case of `simplify` in the `unsupported-predicate` test it prints:
```
[Info#booster] Simplifying predicates
[Warn#booster] Simplify: ignoring unsupported predicates in input
[ErrorDetails#booster] #Not{SortGeneratedTopCell()}(#And{SortGeneratedTopCell()}(#Ceil{SortInt(),SortGeneratedTopCell()}(#DV{SortInt()}(0)),#Equals{SortInt(),SortGeneratedTopCell()}(VarX:SortInt(),#DV{SortInt()}(0))))
```
1. We should probably upstream the `prettyPattern` function.
2. I wasn't sure whether we should print parseable indented kore format or this slightly simpler one in single lines.

geo2a: When eyeballing the logs, it is unlikely that we would benefit from more information than a couple of top-level constructors. I think a single line is better, provided that we have a way to pretty-print it for deeper inspection.

library/Booster/Syntax/Json/Internalise.hs
goodlyrottenapple: we wouldnt need demorgan... #Not ( X #And Y ) ~> notBool (X' andBool Y') if X, Y are bool predicates

unit-tests/Test/Booster/Fixture.hs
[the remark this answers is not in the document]
goodlyrottenapple: fair nough

library/Booster/Syntax/Json/Internalise.hs
```diff
@@ -269,41 +311,111 @@ internaliseTerm ::
...
+
+internalisePred ::
     Flag "alias" ->
     Flag "subsorts" ->
     Maybe [Syntax.Id] ->
     KoreDefinition ->
     Syntax.KorePattern ->
-    Except PatternError InternalisedPredicate
-internalisePredicate allowAlias checkSubsorts sortVars definition@KoreDefinition{sorts} pat = case pat of
+    Except PatternError [InternalisedPredicate]
```
goodlyrottenapple, line 397: not sure if we need to return a list here. since we don't handle #and under a #not and we always explode ands before calling this function (afaik), this always seems to return a singleton

geo2a (approved):

library/Booster/Syntax/Json/Internalise.hs
jberthold: The case of `#And` requires the list, and we need it in the recursion under `#Not`.
Of course we could try to call `explodeAnd` before the recursion and re-assemble the (singleton) results afterwards in the right way (then assuming that we never find `#And` at the top in `internalisePred`).
I can maybe make a follow-up PR to change this (and handle  some more cases inside `#Not` as commented below).

PR #408, opened 2023-12-05
Georgy/fix rule unification

geo2a:
Closes #406 

* Make result indeterminate when trying to bind an already bound variable with a different expression. This was returning a failing result before, causing a soundness bug, because in some cases rules with duplicate variables in LHS would not be applied. 
* Make result indeterminate when unifying an injection with a variable. This allows us to get a clear remainder instead of a weird swapped substitution item
* Only report violating substitution items

PR #420, opened 2023-12-13
Handle conflicting `DomainValue`s in `bindVariable`

geo2a:
In #408, we've made conflicting variable assignment during unification indeterminate rather than failing. However, we can easily handle the case of conflicting domain values.

The motivation for this is the `foundry.prank` rule in Kontrol, which has a higher priority that the opcode fetching rule, and will be attempted at every EVM step, if `<prank>` is set to `true`. This will lead to Booster aborting every EVM step due to a conflicting assignment to the `CD` variable in the  `foundry.prank` rule.

### DIFF
--- a/library/Booster/Pattern/Unify.hs
+++ b/library/Booster/Pattern/Unify.hs
@@ -577,6 +577,7 @@ bindVariable var term = do
             case Map.lookup var currentSubst of
                 Just oldTerm
                     | oldTerm == term -> pure () -- already bound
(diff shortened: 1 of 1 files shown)


goodlyrottenapple (approved):

jberthold (approved):
LGTM.
We could maybe directly `failWith $ DifferentValues ...`, too? (basically inlining the code that unifies two `DomainValue`s)

PR #427, opened 2023-12-15
Concurrent `kore-rpc-booster`

### LINKED ISSUE #383: Support multiple clients in parallel

goodlyrottenapple:
Fixes #383


library/Booster/JsonRpc.hs
```diff
@@ -155,14 +159,40 @@ respond stateVar =
...
+                                        | otherwise ->
+                                            -- this module has previously been added with name-as-id: true
+                                            -- we can allow this, since the contents of the named module
+                                            -- are the same
+                                            pure ()
+                                    _ -> pure ()
+                     in case runExcept
+                            (checkModule >> addToDefinitions newModule{ParsedModule.name = Id moduleHash} state.definitions) of
                             Left err ->
                                 abortWith $ RpcError.backendError RpcError.CouldNotVerifyPattern err
```
jberthold, line 182: We should use the new `DuplicateModuleName` here. Which means 
1. a error type other than `AddModuleError` to throw above (probably just `DuplicateModule`)
2. a separate match to return `DuplicateModuleName` here
```suggestion
                            Left (DuplicateModule name) ->
                                abortWith $ RpcError.backendError RpcError.DuplicateModuleName name
                            Left err ->
                                abortWith $ RpcError.backendError RpcError.CouldNotVerifyPattern err
```

library/Booster/JsonRpc.hs
```diff
@@ -155,14 +159,40 @@ respond stateVar =
...
+                            when nameAsId $
+                                case (Map.lookup (getId $ newModule.name) state.definitions, Map.lookup moduleHash state.definitions) of
+                                    (Just{}, Nothing) ->
+                                        -- another module with the same name already exists
+                                        throwE . AddModuleError $ "Duplicate Module Name"
+                                    (Just nmMod, Just idMod)
+                                        | nmMod /= idMod ->
+                                            -- this module has previously been added and different
+                                            -- module with the same name also already exists
+                                            throwE . AddModuleError $ "Duplicate Module Name"
```
jberthold, line 172: To return `DuplicateModuleName` to the client, we should use `DuplicateModule` here.
```suggestion
                                        throwE $ DuplicateModule $ getId newModule.name
                                    (Just nmMod, Just idMod)
                                        | nmMod /= idMod ->
                                            -- this module has previously been added and different
                                            -- module with the same name also already exists
                                            throwE $ DuplicateModule $ getId newModule.name
```

tools/rpc-client/RpcClient.hs
```diff
@@ -574,19 +575,26 @@ prepareRequestData (Exec file) mbOptFile opts =
 prepareRequestData (Simpl file) mbOptFile opts =
     liftIO $ prepareOneTermRequest "simplify" file mbOptFile opts
 prepareRequestData (AddModule file) mbOptFile opts = do
-    unless (isNothing mbOptFile) $
-        logWarn_ "Add-module mode, ignoring given option file"
     unless (null opts) $
         logWarn_ "Raw mode, ignoring given request options"
```
jberthold, line 579: I think the `opts` are added below?

```suggestion
```

geo2a (approved):

PR #429, opened 2023-12-15
Make ceil generation of in_keys more robust

goodlyrottenapple:
We were previously making the assumption that the only in_keys functions are those for the default built-in maps or cell maps. However, @virgil-serbanuta has a project which defines custom maps with a custom in_keys hook. In order to find the correct in_key symbol, we need to extract the "hook" attribute when internalising symbols and build a rverse lookup map for the given key and map sorts


jberthold (approved):
Should revisit this later but needs to be fixed quickly now

PR #436, opened 2023-12-19
(Dev.) Options to control fall-back reasons and post-exec simplification

### LINKED ISSUE #434: Allow user to control fallback reasons on the CLI
See: https://runtimeverification.slack.com/archives/CC360GUTG/p1702579255239909

I would like to be able to tell the Booster to _not_ fallback to the old backend on branches (to confirm the branch), so that pyk can first apply some semantics specific logic which may mean that the branch just doesn't even need to be considered. In particular, when there is a semantic `terminal` state (not a _stuck_ state), which pyk can identify using semantics-specific information, in KEVM we always get a 5-way branch as the next states. This branch does not need to be analyzed for correctness by the old backend, because we simply don't need to explore past that point.

So instead, if we can have the booster just return branchces immediately, then pyk can decide first (i) is this a branch even worth inspecting, and if so (ii) invoke the `simplify` endpoint directly on the post-branch states anyway. Then we can log in pyk the various simplifications that are actually being applied.

In particular, a CLI option for telling the booster to not fallback on branches would be good. We could also add the other fallback reasons as switches, so that we can experiment directly in pyk with turning those off (to understand them better).

jberthold:
Fixes #434 

* `--fallback-on REASON1,REASON2,...` allows users to set a custom set of reasons on which a fall-back to kore-rpc will be performed. The default is `Branching,Stuck,Aborted` (behaviour as before).
* `--no-post-exec-simplify` switches off the post-exec simplification with kore-rpc (booster results will always be simplified with booster's simplifier before returning).

Both options are meant for developer experimentation. In particular, it is probably not sensible to not have `Aborted` in the `fallback-on` set of halt reasons.

### DIFF
--- a/tools/booster/Proxy.hs
+++ b/tools/booster/Proxy.hs
@@ -430,14 +430,14 @@ respondEither cfg@ProxyConfig{statsVar, boosterState} booster kore req = case re
     postExecSimplify logSettings start mbModule def
         | not cfg.simplifyAtEnd = pure
         | otherwise = \case
-        Execute res ->
-            Execute
-                <$> ( simplifyResult res
-                        `catch` ( \(err :: DecidePredicateUnknown) ->
-                                    pure res{reason = Aborted, unknownPredicate = Just . externaliseDecidePredicateUnknown $ err}
-                                )
-                    )
-        other -> pure other
+            Execute res ->
+                Execute
+                    <$> ( simplifyResult res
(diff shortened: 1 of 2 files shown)


geo2a (approved):
@jberthold I've allowed myself to push a commit that changes how the default abort reasons are displayed in `--help`, so that they look the same as the option format.

PR #438, opened 2023-12-20
431 handle raw term from llvm

### LINKED ISSUE #431: Handle `rawTerm` wrapper from the simplifier
The LLVM backend has added a new way of handling non-symbol terms across API boundaries; previously when doing so we added a spurious injection into `KItem` to non-symbol terms when returning them to callers. This could be worked around, but was not completely sound.

Clients that call backend binding APIs that use the new `rawTerm` symbol must be aware of it when deserializing or parsing binary or textual KORE. For the Python bindings, the deserialization logic is implemented [here directly in C++.](https://github.com/runtimeverification/llvm-backend/blob/7b6d8526dd6c68e06fb8ef3ffd0d69978928d9bc/include/kllvm/binary/deserializer.h#L276-L281)

The intended semantics are to perform the following rewrite unconditionally during parsing or deserialization:
```
rawTerm{}(inj{S, SortKItem{}}(X)) → X
```
where the sort variable `S` is the real sort of the term `X`.

Blocking: https://github.com/runtimeverification/llvm-backend/pull/931

Related:
* https://github.com/runtimeverification/llvm-backend/pull/907
* https://github.com/runtimeverification/llvm-backend/pull/905

jberthold:
Fixes #431 

During deserialisation of terms received from the LLVM backend, transform `rawTerm(inj{S, SortKItem}(X))` into a plain `X` (of expected sort `S`, checked during the transformation).

This code is added so that we can handle both the previous behaviour (of adding an injection `inj{SortX , SortKItem}(X)` to all simplified terms `X:SortX`) and the new behaviour of decorating the former with a `rawTerm(_)`.

A follow-up PR will remove the code that previously handled the `SortKItem` injection.


Baltoli (approved):
My Haskell is pretty rusty but this looks sensible to me!

PR #447, opened 2024-01-09
Internalisation checks for symbol application

### LINKED ISSUE #443: Revisit argument sort and arity errors when internalising symbol applications
Currently, some kinds of errors related to symbol applications are not caught during internalisation:
* wrong arity for the symbol
* wrong argument sort

Checks for these were once in place but must have been lost due to refactorings.

goodlyrottenapple:
fixes #443 by checking each symbol application has the right number of arguments of the correct sort.

### DIFF
--- a/library/Booster/Pattern/Util.hs
+++ b/library/Booster/Pattern/Util.hs
@@ -3,7 +3,7 @@ Copyright   : (c) Runtime Verification, 2022
 License     : BSD-3-Clause
 -}
 module Booster.Pattern.Util (
-    applySubst,
+    substituteInSort,
     sortOfTerm,
     sortOfPattern,
     substituteInTerm,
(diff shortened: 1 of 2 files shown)


test/rpc-integration/test-subsorts/state-not-subsort.execute
```diff
@@ -60,6 +60,22 @@
...
+                "tag": "EVar",
+                "name": "X",
+                "sort": {
+                  "tag": "SortApp",
+                  "name": "SortInt",
+                  "args": []
+                }
+              }
+            ]
+          },
```
jberthold, line 78: Interesting... 
The [configuration in `diamond/test.k`](https://github.com/runtimeverification/hs-backend-booster/blob/53a66a27638834939612e540564f63be90d13079/test/rpc-integration/resources/diamond/test.k) (which is used for this test) is `<k> $PGM:State ~> .k </k><int> 42 </int>, therefore `generatedTop` takes 3 arguments.
So this input actually has _two_ sort errors now (wrong injection within `isInt` in the K-cell, wrong argument sort here), but only the first one is reported (expected response "`KItem` is not subsort of `K`")
Maybe we should 
* separately test the different error cases (wrong injection, wrong argument count, wrong argument sort)
* use `subsort.k` from the internalisation tests for this rpc test? (but probably not necessary)

goodlyrottenapple: I've added two new tests to exercise each of the new checks introduced

geo2a (approved):

jberthold (approved):
LGTM, thanks for adding the new tests!

goodlyrottenapple:
Checked and this does not break any existing kevm/kontrol tests

PR #454, opened 2024-01-10
Documentation of the booster architecture

goodlyrottenapple:


docs/booster.md
```diff
@@ -0,0 +1,231 @@
...
... it can. Once it reaches a point where unification fails due to some part of the configuration containing un-evaluated functions which need to match a concrete value in the rule, the booster will run the simplifier on the current configuration to try to evaluate any function calls inside the configuration before trying to re-write again. This strategy avoids the costly simplifier running after each rewrite, but can be a double edged sword as the booster often ends up building a big stack of unevaluated functions in parts of the configuration. This usually happens in rules operating on the stack/heap/memory cells of a particular semantics, which often functionally update the state from previous configuration by calling an append/upsert/etc. function. In certain instances, the configuration ends up growing at an enormous rate as these thunks build up and can cause failure when the simplifier runs out of memory, trying to simplify such a huge configuration term. To prevent this from happening, the booster proxy contains a simple strategy which allows the user to force the simplification phase to be triggered every _n_ rewrite steps  (_n_ can be specified via a flag at server startup).
```
geo2a, line 30:
```suggestion
The execution can roughly be divided into two phases. The rewrite rule application phase and the simplification phase. Unlike the old backend, which is much more agressive in applying simplification and function rules, the booster tries to apply rewrite rules as long as it can. Once it reaches a point where unification fails due to some part of the configuration containing un-evaluated functions which need to match a concrete value in the rule, the booster will run the simplifier on the current configuration to try to evaluate any function calls inside the configuration before trying to re-write again. This strategy avoids the costly simplifier running after each rewrite, but can be a double edged sword as the booster often ends up building a big stack of unevaluated functions in parts of the configuration. This usually happens in rules operating on the stack/heap/memory cells of a particular semantics, which often functionally update the state from previous configuration by calling an append/insert/etc. function. In certain instances, the configuration ends up growing at an enormous rate as these thunks build up and can cause failure when the simplifier runs out of memory, trying to simplify such a huge configuration term. To prevent this from happening, the booster proxy contains a simple strategy which allows the user to force the simplification phase to be triggered every _n_ rewrite steps  (_n_ can be specified via a flag at server startup).
```

docs/booster.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Simplification / function evaluation
+
+As mentioned in the previous section, the simplification code path is used at two different points of the execution, as well as being exported as a separate `simplify` JSON RPC endpoint. The simplification procedure underpinning all of these use-cases is largely the same and comprises of two main simplifiers, the concrete and symbolic one. Concrete function evaluation is handled by the LLVM backend and thus requires the semantics to be written in such a way, so as to be able to build both the kore definition used by the haskell backend, as well as the LLVM kore definition (should I mention detaisl here??). The booster relies on the LLVM version of a semantics, compiled as a dynamic library, which is loaded when the server starts. During simplification, the term is traversed bottom up and any concrete sub-terms are sent to he LLVM backend to be evaluated.
```
geo2a, line 222: I think the level of detail is good here.

docs/booster.md
```diff
@@ -0,0 +1,231 @@
...
+
+If all the requires clauses have been successfully proven, the booster proceeds to check all the nesures clauses individually, in case any evaluate to false. It then calls the SMT solver again with the full set of current configuration constraints, also adding the new ensures clauses. If this new set of constraints is false/`_|_`, then the whole configuration is `_|_` and the booster records this rewrite as being trivial.
+
+### Simplification / function evaluation
```
geo2a, line 220: After reading this section, I'm left asking for more. I want this section to flow similarly to the rewriting section:
* we describe how we select which equations to attempt
* we describe in detail the process of application of a single equation, perhaps with a diagram

What I also want to convey is this section is the encoding of function rules and simplifications.

geo2a (approved):
Thanks @goodlyrottenapple for this initiative, I think it's a great start! 

I've passed through the document and left several notes about how I see we could iterate on the structure and contents. I suggest we let @jberthold have the draft reviewed too and then merge it. I'd rather submit my catch of typos as a PR.

docs/booster.md
```diff
@@ -0,0 +1,195 @@
...
+│              ┌──────────────┼─────────────┐               │           │ │                   │                            │                        │
+│              │              │             │               │           │ │                   │                   │        │                        │
+│              ▼              ▼             ▼               │           │ │                   │                   │        └───────┐                │
+│                                                           │           │ │                   │                   │                ▼                │
+│          Does not     Simplified      Simplifies          │  ┌────────┼─┼───────────────────┘                   │                                 │
+│          simplify      already                            │  │        │ │                                       │          PS simplify to         │
+│                                                           │  │        │ │                                       │                PS'              │
```
ehildenb, line 77: What's the difference between "Does not simplify" and "Simplified already"?

goodlyrottenapple: Simplified already means that we have done a loop where no rules applied, then we were able to simplify the configuration and tried rule application again, which produced no results. Does not simplify just means we took a rewrite step, then tried to take another one but no applicable rules were found, then we tried to simplify but the configuration wasn't simplified so we return stuck immediately.

I think in the diagram, we can remove this distinction, because functionally, the simplified diagram should be the same

PR #458, opened 2024-01-15
Refactor equals k to follow hook structure

jberthold:
Behaviour has changed:
If the hook cannot evaluate the entire expression, the _original_ term will be returned. Previously, a simplified version would be returned. For instance, for a simplification of expression `Constr(A, 42) ==K Constr(B, 42)`, the old version returned `A ==K B`, whereas the new version returns the original term `Constr(A, 42) ==K Constr(B, 42)`.

### DIFF
--- a/library/Booster/Builtin.hs
+++ b/library/Booster/Builtin.hs
@@ -43,6 +43,8 @@ builtinsKEQUAL :: Map Text BuiltinFunction
 builtinsKEQUAL =
     Map.fromList
         [ "KEQUAL.ite" ~~> iteHook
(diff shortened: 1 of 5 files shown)


jberthold:
No measurable change  in `kevm` test proofs (`PYTEST_PARALLEL=3`)

geo2a (approved):
I like these changes! 

I'm not quite happy with how the unit tests read, i.e it is not always clear what is the expected output in all cases. However. the way we write these test  cases has not been introduced in this PR, and I do not have a clear way towards improvement.

Approving.

PR #461, opened 2024-01-15
Consolidate variable marking API

### LINKED ISSUE #456: Move hadling of variable provenance prefixes to `Booster.Pattern.Util`
When internalizing rules, we rename the rule variables by adding a `Rule#` or `Ex#` prefix to track their provenance and avoid clashing with the variables from the configuration.  As @jberthold suggest in his review for #453, we should move the handling of those prefix literals into `Booster.Pattern.Util` and only expose `VarName -> VarName` functions:
```
markAsExVar, markAsRuleVar, stripMarker :: VarName -> VarName
```

This way we will be able to refactor the variable type more easily if we need to.

geo2a:
Closes #456


jberthold (approved):
LGTM

PR #462, opened 2024-01-15
KEVM abort analysis script

### LINKED ISSUE #460: Automate abort-reason analysis on kevm and kontrol test runs
Current (manual) procedure:

<details><summary>produce-bug-reports script</summary>

   Basically the `performance-test-kevm.sh` only running with `master` and with `--bug-report` options

```
#!/usr/bin/env bash
set -euxo pipefail

# Disable the Python keyring, otherwise poetry sometimes asks for password. See
#  https://github.com/pypa/pip/issues/7883
export PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring

KEVM_VERSION=${KEVM_VERSION:-'master'}

SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"

FEATURE_BRANCH_NAME=${FEATURE_BRANCH_NAME:-"$(git rev-parse --abbrev-ref HEAD)"}
FEATURE_BRANCH_NAME="${FEATURE_BRANCH_NAME//\//-}"

PYTEST_PARALLEL=${PYTEST_PARALLEL:-2}

if [[ $FEATURE_BRANCH_NAME == "master" ]]; then
  FEATURE_BRANCH_NAME="feature"
fi

# Create a temporary directory and store its name in a variable.
TEMPD=$(mktemp -d)

# Exit if the temp directory wasn't created successfully.
if [ ! -e "$TEMPD" ]; then
    >&2 echo "Failed to create temp directory"
    exit 1
fi

# Make sure the temp directory gets removed and kore-rpc-booster gets killed on script exit.
trap "exit 1"           HUP INT PIPE QUIT TERM
trap 'rm -rf "$TEMPD" && killall kore-rpc-booster || echo "No zombie processes found"'  EXIT

feature_shell() {
  GC_DONT_GC=1 nix develop . --extra-experimental-features 'nix-command flakes' --override-input k-framework/booster-backend $SCRIPT_DIR/../ --command bash -c "$1"
}

cd $TEMPD
git clone --depth 1 --branch $KEVM_VERSION https://github.com/runtimeverification/evm-semantics.git
cd evm-semantics

if [[ $KEVM_VERSION == "master" ]]; then
  KEVM_VERSION=$(git name-rev --tags --name-only $(git rev-parse HEAD))
else
  KEVM_VERSION="${KEVM_VERSION//\//-}"
fi

git submodule update --init --recursive --depth 1 kevm-pyk/src/kevm_pyk/kproj/plugin


feature_shell "make poetry && poetry run -C kevm-pyk -- kdist --verbose build evm-semantics.plugin evm-semantics.haskell --jobs 4"

mkdir -p $SCRIPT_DIR/logs
mkdir -p $SCRIPT_DIR/bug-reports

feature_shell "make test-prove-pyk PYTEST_PARALLEL=$PYTEST_PARALLEL PYTEST_ARGS='--maxfail=0 --timeout 7200 -vv --use-booster --bug-report --bug-report-dir $SCRIPT_DIR/bug-reports/kevm-$KEVM_VERSION-$FEATURE_BRANCH_NAME' | tee $SCRIPT_DIR/logs/kevm-$KEVM_VERSION-$FEATURE_BRANCH_NAME.log"
killall kore-rpc-booster || echo "No zombie processes found"
```

</details>

<details><summary>Script to run with a tarball, TODO add llvm-compilation instead of relying on file names and env variable</summary>

```
#! /usr/bin/env bash

# Starts an rpc server and runs a tarball against it using rpc-client
# Environment variables:
#   DEFINITION: kore file to load (optional, otherwise from tarball)
#   LLVM_LIB: default llvm library to load (required)
#   BOOSTER: booster home directory (only relevant if not setting the following)
#   SERVER: kore-rpc server path (default: from build dir in BOOSTER)
#   CLIENT: rpc client path (default: from build dir in BOOSTER)

tarball=${1?"Tarball argument missing"}
shift

tarname=$(basename $tarball)

booster=${BOOSTER:-$(realpath $(dirname $0)/..)}
server=${SERVER:-$booster/.build/booster/bin/kore-rpc-booster}
client=${CLIENT:-$booster/.build/booster/bin/kore-rpc-client}

if [ -z "$DEFINITION" ]; then
    # unpack definition file from tarball, into tmp dir
    mkdir -p tmp
    tar xf $tarball -O definition.kore > tmp/definition.kore
    kore=tmp/definition.kore
else
    kore=$DEFINITION
fi

if [ ! -s $kore ]; then
    echo "Definition file $kore missing/empty"
    exit 1
fi


# find out module name from definition file
MODULE=$(grep -o -e "^module [A-Z0-9-]*" $kore | tail -1 | sed -e "s/module //")

# hard-wire llvm backend, use surplus arguments from the command line as server options
if [ -f ./${tarname}-interpreter.so ]; then
    lib=./${tarname}-interpreter.so
else
    lib=$(realpath ${LLVM_LIB})
fi
if [ ! -f "$lib" ]; then
    echo "A path to an LLVM backend library must be provided in LLVM_LIB"
    exit 2
fi
server_params="--llvm-backend-library $lib $@"

echo "Starting server: $server $kore --module ${MODULE:-UNKNOWN} $server_params"
$server $kore --module ${MODULE?"Unable to find main module"} $server_params > ${tarname}.log 2>&1 &
server_pid=$!

trap 'kill -2 ${server_pid}' ERR EXIT
echo "Server PID ${server_pid}"

sleep 5

echo "Running requests from $tarball against the server"
$client run-tarball $tarball --keep-going

echo "Done with $tarball"
```

</details>

To actually run the tarballs, I used the following command:

```
$ for tar in $(find <location-of-bug-reports> -name \*tar ); do echo "################### $tar ########"; LLVM_LIB=./interpreter.so ../../../scripts/run-with-tarball.sh $tar -l Aborts  --print-stats 2>&1 | tee $(basename $tar).out; done
```
  - Note the `-l Aborts --print-stats` options (which the script passes to the server)

Scripts to produce the statistics: `grep` / `sed` (as shown in the gists), and a simple `Count.hs` script 

* Counting abort reasons:
```
$ grep -e "Uncertain about" *log | sed -e 's,/tmp/tmp.[^/]*/evm-semantics/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/,,' -e "s/.*Uncertain about \(.*\)$/\1/" -e "s/jumpi.true :.*/jumpi.true/" | ../Count
```
* Counting uncertain conditions by rule and actual outcome in kore-rpc:
```
$ grep -E -e "^\[Aborts.*Uncertain about a condition| Booster aborted, kore yields"  *log | sed -n -e '/Uncertain about/,/yields/{/yields/{!p};p}' | sed  -e 's/\(jumpi\.true\) :.*$/\1/' | sed -n -e 'N;s/.*Uncertain about a condition in \(.*\)\n.*kore yields \(.*\)/ \1 -- \2 /g;p' | less | ../Count
```

<details><summary>`Count.hs` script</summary>

```
module Count where

import Data.Bifunctor as B
import Data.Map qualified as Map
import Data.List.Extra
import Data.Ord
import Text.Printf

main = getContents >>=
       mapM_ (uncurry (printf "| %s | %-3d|\n"))
       . sortOn (Down . snd)
       . Map.toList
       .  Map.fromListWith (+)
       . map (,1 :: Int)
       . lines
```

Was always fast enough with plain old `String`s but could also be modified to using `ByteString`.
</details>

goodlyrottenapple:
Fixes #460


scripts/analysis-kevm.sh
```diff
@@ -0,0 +1,40 @@
...
+
+export LOG_DIR="$BUG_REPORT_DIR"
+
+for tar in $(find $BUG_REPORT_DIR -name \*tar ); 
+do 
+  echo "######## $tar ########"; 
+  nix_shell "$SCRIPT_DIR/run-with-tarball.sh $tar -l Aborts --print-stats 2>&1 | tee $BUG_REPORT_DIR/$(basename $tar).out"; 
+done
+
+cd $BUG_REPORT_DIR
```
jberthold, line 34: I would suggest using a separate directory (`$LOG_DIR`?) for the outputs `*.log` and `*.out`, creating one unless an existing one was given by the user.
If we write the output to `$BUG_REPORT_DIR` it will be overwritten by a new run and clean-up is easier if you want to keep the tarballs.

scripts/performance-tests-kevm.sh
```diff
@@ -54,18 +54,45 @@ fi
...
+      ;;
+    -*|--*)
+      echo "Unknown option $1"
+      exit 1
+      ;;
+    *)
+      shift # past argument
+      ;;
+  esac
+done
```
jberthold, line 74: :+1: good idea.
but let's not make it swallow unknown arguments (please remove the `*)` case).
We could move this to the start and make it read `KEVM_VERSION` and `PYTEST_PARALLEL` (which we should default to 3 instead of 3 while at it.

scripts/performance-tests-kevm.sh
```diff
@@ -54,18 +54,45 @@ fi
...
+if [ -z "$BUG_REPORT" ]; then
 if [ ! -e "$SCRIPT_DIR/logs/kevm-$KEVM_VERSION-master-$MASTER_COMMIT_SHORT.log" ]; then
   master_shell "make test-prove-pyk PYTEST_PARALLEL=$PYTEST_PARALLEL PYTEST_ARGS='--maxfail=0 --timeout 7200 -vv --use-booster' | tee $SCRIPT_DIR/logs/kevm-$KEVM_VERSION-master-$MASTER_COMMIT_SHORT.log"
   killall kore-rpc-booster || echo "No zombie processes found"
 fi
 
 cd $SCRIPT_DIR
 python3 compare.py logs/kevm-$KEVM_VERSION-$FEATURE_BRANCH_NAME.log logs/kevm-$KEVM_VERSION-master-$MASTER_COMMIT_SHORT.log > logs/kevm-$KEVM_VERSION-master-$MASTER_COMMIT_SHORT-$FEATURE_BRANCH_NAME-compare
+else
+  cp $KDIST_DIR/evm-semantics/haskell/llvm-library/interpreter.* $SCRIPT_DIR/bug-reports/kevm-$KEVM_VERSION-$FEATURE_BRANCH_NAME/interpreter.dylib
```
jberthold, line 96: Unfortunately this won't work, the different tests need different `interpreter.so/dylib` files (some of the tests use special symbols, like `runLemma`). 
The safe option would be to use `llvm-kompile`  (and `llvm-kompile-matching`) inside `run-with-tarball` for each tarball. Otherwise some tests will crash early (you still get most results)

scripts/run-with-tarball.sh
```diff
@@ -0,0 +1,74 @@
...
+
+trap 'kill -2 ${server_pid}' ERR EXIT
+echo "Server PID ${server_pid}"
+
+i=15
+while ! lsof -a -p${server_pid} -sTCP:LISTEN -iTCP && [[ $i -ge 0 ]]; do
+    echo "Waiting for server ($i attempts left)"
+    i=$((i-1))
+    sleep 1
+done
```
jberthold, line 64: The client meanwhile has its own wait loop so we could remove this (and remove `lsof` from the nix flake)

scripts/run-with-tarball.sh
```diff
@@ -0,0 +1,74 @@
...
+# hard-wire llvm backend, use surplus arguments from the command line as server options
+if [ -f ./${tarname}-interpreter.so ]; then
+    lib=./${tarname}-interpreter.so
+else
+    lib=$(realpath ${LLVM_LIB})
+fi
+if [ ! -f "$lib" ]; then
+    echo "A path to an LLVM backend library must be provided in LLVM_LIB"
+    exit 2
+fi
```
jberthold, line 49: I ended up extending my script today to call `llvm-kompile` and `llvm-kompile-matching`  so I could run `kontrol` tarballs better....
We would have to somehow pass in the `PLUGIN_DIR`  variable pointing to the nix store from the other script.

```suggestion
# build llvm backend unless provided
if [ -z "${LLVM_LIB}" ]; then
    mkdir -p tmp
    tar xf $tarball -O llvm_definition/definition.kore > tmp/llvm-definition.kore
    if [ ! -d "${PLUGIN_DIR}" ]; then
        echo "Either LLVM_LIB or PLUGIN_DIR must be provided"
        exit 2
    fi
    #generate matching data
    (cd tmp && mkdir -p dt && llvm-kompile-matching llvm-definition.kore qbaL ./dt 1/2)
    # find library dependencies and source files
    for lib in libff libcryptopp libsecp256k1; do
        LIBFILE=$(find ${PLUGIN_DIR} -name "${lib}.a" | head -1)
        [ -z "$LIBFILE" ] && (echo "[Error] Unable to locate ${lib}.a"; exit 1)
        PLUGIN_LIBS+="$LIBFILE "
        PLUGIN_INCLUDE+="-I$(dirname $LIBFILE)/../include "
    done
    PLUGIN_CPP="${PLUGIN_DIR}/include/plugin-c/blake2.cpp ${PLUGIN_DIR}/include/plugin-c/crypto.cpp ${PLUGIN_DIR}/include/plugin-c/plugin_util.cpp"

    # kompile llvm-definition to interpreter
    case "$OSTYPE" in
        linux*)
            LPROCPS="-lprocps"
            LIBSUFFIX="so"
            ;;
        *)
            LPROCPS=""
            LIBSUFFIX="dylib"
            ;;
    esac
    llvm-kompile tmp/llvm-definition.kore tmp/dt c -- \
             -fPIC -std=c++17 -o tmp/interpreter \
             $PLUGIN_LIBS $PLUGIN_INCLUDE $PLUGIN_CPP \
             -lcrypto -lssl $LPROCPS
    lib=tmp/interpreter.$LIBSUFFIX
else
    lib=$(realpath ${LLVM_LIB})
fi
if [ ! -f "$lib" ]; then
    echo "Problem with LLVM backend library at $lib"
    exit 2
fi
```

scripts/run-with-tarball.sh
goodlyrottenapple: i couldn't get your original code working, so this was copied from the integration tests and is pretty well tested

scripts/performance-tests-kevm.sh
goodlyrottenapple: ah :/ that's unfortunate, but should be able to re-work easily enough

scripts/booster-analysis.sh
```diff
@@ -0,0 +1,59 @@
...
+#   $SCRIPT_DIR/run-with-tarball.sh $tar -l Aborts --print-stats 2>&1 | tee $LOG_DIR/$(basename $tar).out; 
+# done
+
+cd $LOG_DIR
+
+# Counting abort reasons
+grep -e "Uncertain about" *log | sed -e 's,/tmp/tmp.[^/]*/evm-semantics/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/,,' -e "s/.*Uncertain about \(.*\)$/\1/" -e "s/jumpi.true :.*/jumpi.true/" | nix_shell "runghc $SCRIPT_DIR/Count.hs" > abort_reasons.count
+
+# Counting uncertain conditions by rule and actual outcome in kore-rpc
+grep -E -e "^\[Aborts.*Uncertain about a condition| Booster aborted, kore yields"  *log | sed -n -e '/Uncertain about/,/yields/{/yields/{!p};p}' | sed  -e 's/\(jumpi\.true\) :.*$/\1/' | sed -n -e 'N;s/.*Uncertain about a condition in \(.*\)\n.*kore yields \(.*\)/ \1 -- \2 /g;p' | less | nix_shell "runghc $SCRIPT_DIR/Count.hs" > uncertain_conditions.count
```
jberthold, line 59:
```suggestion
grep -e "Uncertain about" *log | sed -e 's,/tmp/tmp.[^/]*/evm-semantics/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/,,' -e "s/.*Uncertain about \(.*\)$/\1/" -e "s/jumpi.true :.*/jumpi.true/" | nix_shell "runghc --ghc-arg='--package extra' $SCRIPT_DIR/Count.hs" > abort_reasons.count

# Counting uncertain conditions by rule and actual outcome in kore-rpc
grep -E -e "^\[Aborts.*Uncertain about a condition| Booster aborted, kore yields"  *log | sed -n -e '/Uncertain about/,/yields/{/yields/{!p};p}' | sed  -e 's/\(jumpi\.true\) :.*$/\1/' | sed -n -e 'N;s/.*Uncertain about a condition in \(.*\)\n.*kore yields \(.*\)/ \1 -- \2 /g;p' | less | nix_shell "runghc --ghc-arg='--package extra' $SCRIPT_DIR/Count.hs" > uncertain_conditions.count
```

scripts/booster-analysis.sh
```diff
@@ -0,0 +1,59 @@
...
+
+run_tarball(){
+  echo "######## $1 ########"; 
+  $SCRIPT_DIR/run-with-tarball.sh $1 -l Aborts --print-stats 2>&1 | tee $LOG_DIR/$(basename $1).out; 
+}
+
+export -f run_tarball
+export SCRIPT_DIR
+
+find $BUG_REPORT_DIR -name \*tar -print0 | xargs -0 -t -I {} -P $(nproc) bash -c 'run_tarball "$@"' _ {}
```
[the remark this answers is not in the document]
goodlyrottenapple: We dont care about the timings here so i would definitely stick with nproc....could make it a parameter defaulting to nproc maybe?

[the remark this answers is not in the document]
jberthold: I'd like to have the control, but defailting to `nproc` is fine with me.

jberthold (approved):

PR #464, opened 2024-01-16
Report rewrite as stuck when term index is none

### LINKED ISSUE #457: Booster should return `"stuck"` (not `"aborted"`) when no rules apply
Currently, `booster-dev` returns `"aborted"` on several integration test inputs where no rules apply, for example:

* [test-questionmark-vars/response-one-ques.booster-dev](https://github.com/runtimeverification/hs-backend-booster/blob/main/test/rpc-integration/test-questionmark-vars/response-one-ques.booster-dev)
* [test-compute-ceil/response-evaluate-g.booster-dev](https://github.com/runtimeverification/hs-backend-booster/blob/main/test/rpc-integration/test-compute-ceil/response-evaluate-g.booster-dev)

and perhaps others. If these test had been ran with `kore-rpc-booster`, we would get `"stuck"`, since we would pass the result through Kore too. We need to return `"aborted"` from Booster directly, thus saving an attempt to apply rules in Kore.

geo2a:
Closes #457

When a term's index is `None`, the terms is stuck.

Previously, `rewriteStep` threw an exception when a term's index was `None`, causing an `RewriteAborted` result to be returned by `performRewrite`, leading to `Aborted` execute response.

### DIFF
--- a/library/Booster/Pattern/Rewrite.hs
+++ b/library/Booster/Pattern/Rewrite.hs
@@ -98,22 +98,23 @@ rewriteStep ::
     Pattern ->
     RewriteT io (RewriteFailed "Rewrite") (RewriteResult Pattern)
 rewriteStep cutLabels terminalLabels pat = do
-    let termIdx = kCellTermIndex pat.term
-    when (termIdx == None) $ throw (TermIndexIsNone pat.term)
     def <- getDefinition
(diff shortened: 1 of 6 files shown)


geo2a:
Subsumed by https://github.com/runtimeverification/hs-backend-booster/pull/465

PR #470, opened 2024-01-17
Don't lose traces in Booster

### LINKED ISSUE #466: Backend not reporting all the successfully applied rules in the logs
Inspired by the work here: https://github.com/runtimeverification/pyk/pull/817. We're trying to incorporate log output now in the pyk KCFG viewer.

Attached is a bug report from a Kontrol execution where the logs produced don't seem to agree with the depth count. Fro example, in the attached bug report, we turn on `log-successful-rewrites` in request 006, and it returns saying that it executed to depth 341 due to branching, but the `logs` field only contains a single rewrite rule applied (where I would expect 341 rewrite rules to be listed).

[bad-logs.tar.gz](https://github.com/runtimeverification/hs-backend-booster/files/13954063/bad-logs.tar.gz)

geo2a:
Fixes #466 

Use `catMaybes`, not `mapM` on a list of `Maybe`s

Todo:
~- [ ] Add integration tests for combinations of tracing options~ postpone to follow-up PR


ehildenb:
Confirmed that this seems to work on the handful of examples I tried in KEVM!

ehildenb:
See also: https://github.com/runtimeverification/hs-backend-booster/issues/472

goodlyrottenapple (approved):

library/Booster/JsonRpc.hs
```diff
@@ -580,21 +580,20 @@ execResponse mbDuration req (d, traces, rr) originalSubstitution unsupported = c
...
          in case (timingLog, traceLogs) of
-                (Nothing, Nothing) -> Nothing
-                (Nothing, Just []) -> Nothing
-                (Nothing, Just xs@(_ : _)) -> Just xs
-                (Just t, Nothing) -> Just [t]
-                (Just t, Just xs) -> Just (t : xs)
+                (Nothing, []) -> Nothing
+                (Nothing, xs@(_ : _)) -> Just xs
+                (Just t, []) -> Just [t]
+                (Just t, xs) -> Just (t : xs)
```
goodlyrottenapple, line 596: these two cases could be merged

PR #474, opened 2024-01-22
[DO NOT MERGE YET] Revert "Update dependency: deps/haskell-backend_release"

goodlyrottenapple:
Reverts runtimeverification/hs-backend-booster#450


Scott-Guest:
Don't merge this! Going to just manually push all the KRYPTO changes through right now to unblock the booster

PR #478, opened 2024-01-25
Add build time to version info

geo2a:
This RP modifies the output of `kore-rpc-booster --version` to include the build time:

```
kore-rpc-booster --version
hs-backend-booster version:
  revision:	0f1143648471ab3ea61da49b7d8fff3fa67168a0 (dirty)
  branch:	georgy/better-unclear-condition
  last commit:	Wed Jan 24 10:10:46 2024 +0000
  build time:	2024-01-25 08:24:28.060971253 UTC
```
My primary motivation for this change is convincing myself that I'm not using a stale build when testing locally.


goodlyrottenapple (commented):
this probably wont work in nix/make the build non-deterministic, so may want to disable this when building via nix

geo2a:
Makes sense. Let's close it and I can reapply it locally if I'm paranoid again.

PR #482, opened 2024-01-30
Make connection attempts an env variable

geo2a:
In `test/rpc-integration/runDirectoryTest.sh` we wait for the server for 15 seconds. This PR makes this value configurable with an environment variable, keeping 15 as default.


jberthold (approved):
LGTM.
BTW the `RpcClient` has [its own retry mechanism](https://github.com/runtimeverification/hs-backend-booster/blob/main/tools/rpc-client/RpcClient.hs#L162-L177), we could expose its parameters as program options.

PR #484, opened 2024-01-30
Remove states from RPC logs, include only rule unique IDs

geo2a:
This PR removes the states from Booster's rewrite and equation traces. Only rule unique IDs are included in the RPC logs now.

Here's a quick performance comparison on the execute request `test-fountry-bug-report/state-007.send` with `log-successful-rewrites` and `log-successful-simplifications` set to `true`:

| Git revision                          | Time       | Comment                                            |
|:--------------------------------------|:-----------|----------------------------------------------------|
| georgy/rpc-logs-only-labels           | 3.286555 s | this PR                                            |
| main (0ef6ecd)                        | 3.259145294| Georgy's log fix reverted                          |
| 0096f93                               | 35s        | logs with states, `-N` still there                 |

The performance with unique IDs only seems to be on par with not collecting traces at all.

I have also checked that:

- [x] we still emit rewrite logs with states to `stderr` when `-l Rewrite/RewriteSuccess` is passed and `log-successful/failed-rewrites: false`
- [x] we still emit simplification logs with states to `stderr` when `-l Simplify/SimplifySuccess` is passed and `log-successful/failed-simplifications: false `

# Summary of changes

The idea is to stop accumulating states in `RewriteTrace`s and `EquationTrace`s while still keeping the possibility to emit them in `stderr` logs and possibly stream them to the client in future.

## Refactoring of `EquationTrace`

* the `EquationMetadata` type is introduced to hold an equation's location, rule label and unique id information
* `EquationTrace` is split into two constructors, `EquationApplied` and `EquationNotApplied`, where the former can hold the post-term, and the latter an `ApplyEquationResult`
* `EquationTrace` is made parametric in the term type, similar to `RewriteTrace`. The old `EquationTrace` now becomes `EquationTrace Term`
* a function `eraseStates :: EquationTrace Term -> EquationTrace ()` is added to remove the terms from a trace, leaving only metadata. This function will be called in the rewriting code to erase the states from the equation traces.

## Refactoring of `RewriteTrace`

* The `RewriteSimplified` data constructor is changed to accommodate the parametricity of `EquationTrace`. Since `EquationTrace` usually holds a `Term`, rather than a `Pattern`, we need the `TracePayload` type family to connect the two. The intention is to use `RewriteTrace ()` everywhere, and all the parametricity is only needed to make the code changes minimal. Once we are happy with the API, we could remove parameters from both `RewriteTrace` and `EquationTrace`.
* As with `EquationTrace`, a we now have the `eraseStates :: RewriteTrace Pattern -> RewriteTrace ()` function that drops the patters from the traces. It also erases the states in euqational traces stored in `RewriteSimplified` constructors.

# Future work

The `ApplyEquationResult` holds one success and several error cases: 

```haskell
data ApplyEquationResult
    = Success Term
    | FailedMatch MatchFailReason
    | IndeterminateMatch
    | IndeterminateCondition [Predicate]
    | ConditionFalse Predicate
    | EnsuresFalse Predicate
    | RuleNotPreservingDefinedness
    | MatchConstraintViolated Constrained VarName
```

I think it would be cleaner to factor out the error cases into a separate type:

```haskell
data ApplyEquationFailureReason
    = FailedMatch MatchFailReason
    | IndeterminateMatch
    | IndeterminateCondition [Predicate]
    | ConditionFalse Predicate
    | EnsuresFalse Predicate
    | RuleNotPreservingDefinedness
    | MatchConstraintViolated Constrained VarName

data ApplyEquationResult
    = ApplyEquationSuccess Term
    | ApplyEquationFailure ApplyEquationFailureReason
```

This would make `EquationTrace` cleaner, as currently it can have represent absurd case of `EquationNotApplied ... Success`.


geo2a:
## KEVM performance

No difference with `main` where the logs are not emitted at all.

| Test                                    | georgy-rpc-logs-only-labels time | master-0ef6ecd time | (georgy-rpc-logs-only-labels/master-0ef6ecd) time
|-----------------------------------------|----------------------------------|---------------------|---------------------------------------------------
| erc20/hkg/transferFrom-failure-1-spec.k | 34.15                            | 39.95               | 0.8548185231539424        
| erc20/ds/approve-failure-spec.k         | 27.05                            | 28.16               | 0.9605823863636364        
| erc20/ds/approve-success-spec.k         | 27.97                            | 29.08               | 0.9618294360385145        
| erc20/hkg/transfer-success-2-spec.k     | 41.22                            | 35.71               | 1.1542985158218986        
| TOTAL                                   | 130.39                           | 132.9               | 0.9811136192626033

library/Booster/Pattern/Rewrite.hs
```diff
@@ -474,9 +475,23 @@ data RewriteTrace pat
     | -- | attempted rewrite failed
       RewriteStepFailed (RewriteFailed "Rewrite")
     | -- | Applied simplification to the pattern
-      RewriteSimplified [EquationTrace] (Maybe EquationFailure)
-    deriving stock (Eq, Show)
-    deriving (Functor, Foldable, Traversable)
+      RewriteSimplified [EquationTrace (TracePayload pat)] (Maybe EquationFailure)
+
+type family TracePayload pat where
```
goodlyrottenapple, line 480:
```suggestion
type family EquationTracePayload pat where
```

goodlyrottenapple (approved):
LGTM!

PR #486, opened 2024-01-31
Make equation limits configurable

geo2a:
* move equation iteration and recursion limits to `EquationConfig`

### DIFF
--- a/library/Booster/Pattern/ApplyEquations.hs
+++ b/library/Booster/Pattern/ApplyEquations.hs
@@ -124,6 +124,8 @@ data EquationConfig = EquationConfig
     , llvmApi :: Maybe LLVM.API
     , smtSolver :: Maybe SMT.SMTContext
     , doTracing :: Bool
+    , maxIterations :: Int
+    , maxRecursion :: Int
     }
 
 data EquationState = EquationState
@@ -291,34 +293,42 @@ runEquationT doTracing definition llvmApi smtSolver sCache (EquationT m) = do
     (res, endState) <-
         flip runStateT (startState sCache) $
             runExceptT $
-                runReaderT m EquationConfig{definition, llvmApi, smtSolver, doTracing}
+                runReaderT
+                    m
+                    EquationConfig
+                        { definition
+                        , llvmApi
+                        , smtSolver
+                        , doTracing
+                        , maxIterations = 100
+                        , maxRecursion = 5
+                        }
     pure (res, toList endState.trace, endState.cache)
 
 iterateEquations ::
     MonadLoggerIO io =>
-    Int ->
-    Int ->
     Direction ->
     EquationPreference ->
     Term ->
     EquationT io Term
-iterateEquations maxRecursion maxIterations direction preference startTerm = do
+iterateEquations direction preference startTerm = do
     logOther (LevelOther "Simplify") $ "Evaluating " <> renderText (pretty startTerm)
     result <- pushRecursion startTerm >>= checkCounter >> go startTerm <* popRecursion
     logOther (LevelOther "Simplify") $ "Result: " <> renderText (pretty result)
     pure result
   where
-    checkCounter counter
-        | counter > maxRecursion =
+    checkCounter counter = do
+        config <- getConfig
+        when (counter > config.maxRecursion) $
             throw . TooManyRecursions . (.recursionStack) =<< getState
-        | otherwise = pure ()
 
     go :: MonadLoggerIO io => Term -> EquationT io Term
     go currentTerm
         | (getAttributes currentTerm).isEvaluated = pure currentTerm
         | otherwise = do
+            config <- getConfig
             currentCount <- countSteps
-            when (currentCount > maxIterations) $
+            when (currentCount > config.maxIterations) $
                 throw $
                     TooManyIterations currentCount startTerm currentTerm
             pushTerm currentTerm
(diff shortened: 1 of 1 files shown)


library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -318,34 +322,42 @@ runEquationT doTracing definition llvmApi smtSolver sCache (EquationT m) = do
...
-                runReaderT m EquationConfig{definition, llvmApi, smtSolver, doTracing}
+                runReaderT
+                    m
+                    EquationConfig
+                        { definition
+                        , llvmApi
+                        , smtSolver
+                        , doTracing
+                        , maxIterations = UnsafeGlobalState.unsafeReadGlobalMaxIterations
+                        , maxRecursion = UnsafeGlobalState.unsafeReadGlobalMaxRecursion
```
jberthold, line 333: We are actually in `IO` here so we could make the `UnsafeGlobalState` slighly less unsafe (still using a hidden global variable but no `unsafePerformIO`)

library/Booster/UnsafeGlobalState.hs
```diff
@@ -0,0 +1,33 @@
...
+{-# NOINLINE unsafeReadGlobalMaxIterations #-}
+unsafeReadGlobalMaxIterations :: Bound "Iterations"
+unsafeReadGlobalMaxIterations = unsafePerformIO $ readIORef globalMaxIterations
+
+writeGlobalMaxIterations :: Bound "Iterations" -> IO ()
+writeGlobalMaxIterations = atomicWriteIORef globalMaxIterations
+
+{-# NOINLINE globalMaxRecursion #-}
+globalMaxRecursion :: IORef (Bound "Recursion")
+globalMaxRecursion = unsafePerformIO (newIORef 5)
```
jberthold, line 26: Should we maybe use _a single `IORef`_  holding`EquationOptions` ?

o/t https://drive.google.com/drive/folders/1YkPxlz862FhttZ5uVRC3z9FN5Ax7wWkX?usp=sharing

geo2a: Sounds good! Done.

library/Booster/Pattern/ApplyEquations.hs
geo2a: Done.

library/Booster/JsonRpc.hs
```diff
@@ -303,7 +305,7 @@ respond stateVar =
                         Log.logOtherNS "booster" (Log.LevelOther "Simplify") $ renderText (pretty ps)
                         let predicates = map (substituteInPredicate ps.substitution) $ Set.toList ps.boolPredicates
                         ApplyEquations.simplifyConstraints
-                            doTracing
+                            (coerce doTracing)
```
goodlyrottenapple, line 308: one nitpick I would have with this change would be to construct `doTracing` as a `Flag "CollectEquationTraces"` instead of coercing at all the call sites, which somewhat negates the usefulness of the flag construct. The idea of the flag is to make sure we don't mix up the arguments and pass the wrong flag to a function, hence coercing should be reserved for turning a flag back into a boolean inside a pattern guard or an if statement.

unit-tests/Test/Booster/Pattern/Rewrite.hs
```diff
@@ -269,7 +273,7 @@ failsWith t err =
 runRewrite :: Term -> IO (Natural, RewriteResult Term)
 runRewrite t = do
     (counter, _, res) <-
-        runNoLoggingT $ performRewrite False def Nothing Nothing Nothing [] [] $ Pattern_ t
+        runNoLoggingT $ performRewrite (coerce False) def Nothing Nothing Nothing [] [] $ Pattern_ t
```
goodlyrottenapple, line 276: in a similar vein, perhaps these could be made into pattern synonyms, something like
```
pattern CollectRewriteTraces = Flag True
pattern DontCollectRewriteTraces = Flag False
```

goodlyrottenapple:
@geo2a I've left two comments, the second one is a minor one I don't mind either way, but the first one should perhaps be addressed

library/Booster/JsonRpc.hs
jberthold: Agree, `coerce`ing `Bool -> Flag name` is opposed to the idea of the `Flag`s per se.
Note that `coerce doTracing` appears for both the rewrite and the simplification trace flags, so we would need to either define two flags or use `doTracing :: forall name. Flag name` (which is kind-of unsafe, too).

unit-tests/Test/Booster/Pattern/Rewrite.hs
geo2a: Good idea! I've done that.

library/Booster/CLOptions.hs
```diff
@@ -2,12 +2,14 @@
...
     CLOptions (..),
+    EquationOptions (..),
     clOptionsParser,
     adjustLogLevels,
     versionInfoParser,
 ) where
 
 import Booster.Trace (CustomUserEventType)
+import Booster.Util (Bound (..))
 import Booster.VersionInfo (VersionInfo (..), versionInfo)
```
jberthold, line 13: :thinking: maybe not too important but I usually put the library imports first, and our "own" imports afterwards (separated by a blank line). These imports were obviously not like that before so nvm... :shrug:

jberthold (approved):
LGTM, thanks for the changes

library/Booster/CLOptions.hs
geo2a: yeah, I've noticed, but I though it does not matter that much to drag another change into this PR

goodlyrottenapple (approved):

PR #489, opened 2024-02-01
Shorten simplification warning and error messages

geo2a:
When equation application fails with `TooManyIterations` or another data constructor of `EquationFailure`, we just `show` the data constructor, log it as a warning or return as error response (whichever is appropriate). However, it is usually only the data constructor tag (i.e. `TooManyIterations`) and not its arguments (usually huge terms) that is useful. This PR makes it so we print only the constructor name

**After:** 
```
[Warn#proxy] Problem with simplify request :  Aborted - TooManyRecursions [Term (TermAtt..............................................................................................................goes on forever.............................................
```

**After:** 
```
[Warn#proxy] Problem with simplify request :  Aborted - TooManyRecursions
```


goodlyrottenapple (changes requested):
Why not just `[Warn#proxy] Problem with simplify request :  Aborted - TooManyRecursions` ?

geo2a:
@goodlyrottenapple because we need to then specifically handle every error case and have specific fragile code that prints the constructors.

goodlyrottenapple (approved):

Issue #488, opened 2024-02-01
`dlopen` the LLVM library for each session worker


goodlyrottenapple:
Now that we have multi-threaded runtime, @nwatson22 is reporting crashes when running with multiple threads. The LLVM backend is known not to be thread-safe, so we need to either put a lock on the loaded library or dynamically load multiple copies, one per worker thread.

PR #500, opened 2024-02-09
Concurrent llvm lib

goodlyrottenapple:
Implements #488 

This change allows us to remove the mutex on the llvm library by copying and loading a copy of the llvm library per session. This means if multiple clients connect, they each get a separate version of the library which avoids issues with thread safety. In the long term, it would be good if the library could be made thread safe, as this approach may use up a lot of memory if many clients connect simultaneously.


test/llvm-integration/LLVM.hs
```diff
@@ -59,24 +56,25 @@ kompiledPath = "test/llvm-integration/definition/llvm-kompiled"
 dlPath = kompiledPath </> "interpreter" <.> (if os == "darwin" then ".dylib" else ".so")
 
 main :: IO ()
-main = defaultMain =<< testSpec "LLVM simplification" llvmSpec
```
goodlyrottenapple, line 62: @jberthold I reverted this to what it was before the mutex change... not sure what exactly is the difference, but this was segfaulting again after I removed the mutex.

jberthold: What I think happens is that the test code creates a single loaded DL and runs tests in parallel  for each generated input (when using tasty's `testSpec` - osing `hspec` sequentialises the tests and avoids the race conditions in the library).
You have added the `loadAPI` helper so (in theory) each of the property tests should get its own copy. That would mean it is the parallelism in each single test case that causes the race conditions. 

To test this hypothesis: Does it still segfault if you request just one success for each property? We could also embed the loading into the property itself,  which of course has considerable overhead. It would be good to understand what exactly causes the race condition.

That said, I would suggest we keep the mutex because we know that any concurrent DL access may cause a segfault without it. The test demonstrates that the DL access is unsafe, we'd rather be slow than at risk.

Issue #498, opened 2024-02-08
Booster falls back to the Haskell backend because of unevaluated functions.


virgil-serbanuta:
Versions:
```
$ kompile --version
K version:    v6.1.82-0-g8c99ee7740
Build date:   Vi ian 19 11:28:02 EET 2024
$ kore-rpc-booster --version
hs-backend-booster version:
  revision:     6067fee77da53ac357ce5d17ad34e4639070804a
  branch:       HEAD
  last commit:  Tue Jan 16 09:27:16 2024 -0700
```

Bug report:
[unevaluated-functions.tar.gz](https://github.com/runtimeverification/hs-backend-booster/files/14211022/unevaluated-functions.tar.gz)

Logs:
```
[...]
[Info] Process request 1 execute
[Info#proxy] Starting execute request
[Rewrite] Uncertain about a condition in rule /tmp/kbuild-kmxwasm-semantics-_exg7e37/kmxwasm-semantics/include/kmxwasm-semantics/mx-semantics/vmhooks/bigIntOps.md :  ( 349
, 10 ) :  _andBool_(
    _andBool_(
        _in_keys(_)_MAP_Bool_KItem_Map(
            "4294967195",
            {"4294967282"->"100", "4294967195"->VarARG_0:SortInt{}}
        ),
        isInt(
            kseq(
                _[_]orDefault__MAP_KItem_Map_KItem_KItem(
                    {"4294967282"->"100", "4294967195"->VarARG_0:SortInt{}},
                    "4294967195",
                    "0"
                ),
                dotk()
            )
        )
    ),
    _andBool_(
        _in_keys(_)_MAP_Bool_KItem_Map(
            "4294967282",
            {"4294967282"->"100", "4294967195"->VarARG_0:SortInt{}}
        ),
        isInt(
            kseq(
                _[_]orDefault__MAP_KItem_Map_KItem_KItem(
                    {"4294967282"->"100", "4294967195"->VarARG_0:SortInt{}},
                    "4294967282",
                    "0"
                ),
                dotk()
            )
        )
    )
) @(hs-backend-booster-0.0.0-qJEsy50ZKtfpZmzEpvni:Booster.Pattern.Rewrite library/Booster/Pattern/Rewrite.hs:616:18)
[...]

PR #502, opened 2024-02-13
Stronger map matching

jberthold:
Previously, only fully concrete maps (all keys and values identical) would have matched, which is highly unlikely when matching simplification rules.

Now we compare the key sets for (syntactic) equality after substituting with the current substitution, and then try to perform matching on the remainder if a solution is at all possible.

Most importantly, maps of fully-concrete keys can now be handled, matching the contained values. In addition, the case of matching a single unknown key out of a singleton map (i.e., without opaque rest in the subject map) is now possible.

Any ambiguous case is still rejected as `MatchIndeterminate` but some cases can be decided as `MatchFailed` immediately.

Aims to address problems reported in #498


jberthold:
Superseded by #506

PR #504, opened 2024-02-13
Keep equation traces from recursive evaluation

geo2a:


geo2a:
Subsumed by https://github.com/runtimeverification/hs-backend-booster/pull/516

PR #508, opened 2024-02-15
Fix add-module

goodlyrottenapple:
Fixes a corner case tested for in https://github.com/runtimeverification/pyk/pull/871 where

* we send module `M1` with the name `"m<hash of M2>"` and `"name-as-id": true`
* we send `M2`

In this case, the server should reject `M2`, but currently both requests succeed, with `M1` being overridden by `M2` 

Blocked on https://github.com/runtimeverification/haskell-backend/pull/3721


jberthold:
I added the tests extracted from the `pyk` branch. Now we are probably adding module `A` about 9 times when the test is running (`pyk` tests run in isolation from each other, the ones I made don't... and don't have to IMHO).

geo2a (approved):

PR #510, opened 2024-02-16
Log ceil analysis with `-l Ceil`

goodlyrottenapple:
When starting the server with `-l Ceil`, with this change, we will emit the logs of the ceil analysis of rewrite rules in the given definition file, e.g.:

```
[Ceil#booster] Partial symbols found in rule /Users/sam/git/hs-backend-booster/bug_reports/concurrency_crash/imp.k :  ( 54
, 8 )
  non-total symbol Lbl_modInt_
rule does NOT preserve definedness. Partially computed ceils:
  _=/=Int_(Rule#VarI2:SortInt{}, "0")
  #Ceil( Rule#VarI2:SortInt{} )
```


geo2a (changes requested):
I think it would be best to keep the output concise and only do not print the rule itself. Let's restrict the output to:
* rule location and, if available, label
* partial symbols
* computed ceils --- I think we could also rename it to "inferred definedness conditions"

I'd guess is is not going to be very useful for the potential users of this feature to see the pretty-printed rules. We can of course still leave the possibility to print rules, but I strongly suggest we hide it behind an additional verbosity option.

tools/booster/Server.hs
```diff
@@ -144,17 +146,22 @@ main = do
...
                     flip runLoggingT monadLogger $
                         Logger.logErrorNS "proxy" $
                             "Main module " <> mainModuleName <> " not found in " <> Text.pack definitionFile
                     liftIO exitFailure
 
+                flip runLoggingT monadLogger $
+                    forM_ (Map.elems definitionsWithCeilSummaries) $ \(_, summaries) ->
+                        forM_ summaries $ \summary ->
+                            Logger.logOtherNS "booster" (Logger.LevelOther "Ceil") $ renderText (pretty summary)
+
```
jberthold, line 164: :heavy_check_mark:  This will provide the required information to the users.
However, maybe we could add the logging to the `computeCeilsDefinition` itself and run it with `runStderrLoggingT` (or with the new handle-based version), then we can use the same logging when we extend the analysis later.

jberthold: I don't see a change to the `Pretty ComputeCeilSummary` instance... so I assume the output still includes the rule?

goodlyrottenapple: sorry I thought I pushed my changes yday :/

geo2a (approved):

tools/booster/Server.hs
goodlyrottenapple: @jberthold I double checked and I wrote a separate instance for pretty printing directly into this code as we went to have a more verbose instance when invoking with the parse binary (I think)

geo2a (approved):

PR #512, opened 2024-02-20
Allow sending different log levels to different handles

geo2a:
This PR modifies the logging infrastructure by allowing to redirect logs at different levels to different handles. The functionality is not yet used and the previous behavior of logging everything to `stderr` is preserved.

In a follow-up PR we may introduce logging of some of the levels to a file. To achieve that, we will need to modify the `logLevelToHandle` function in `tools/booster/Server.hs` (for `kore-rpc-booster`) and in `dev-tools/booster-dev/Server.hs` (for `booster-dev`).


goodlyrottenapple (approved):

library/Booster/JsonRpc/Utils.hs
```diff
@@ -260,3 +264,21 @@ diffBy def pat1 pat2 =
...
+    Log.LogLevel ->
+    Log.LogStr ->
+    IO ()
+handleOutput levelToHandle loc src level msg =
+    let bytes = Log.fromLogStr $ Log.defaultLogStr loc src level msg
+     in BS8.hPutStrLn (levelToHandle level) bytes
+
+-- | Run a logging computation, redirecting various levels to the handles specified by the first arguments
+runHandleLoggingT :: (Log.LogLevel -> IO.Handle) -> Log.LoggingT m a -> m a
+runHandleLoggingT = flip Log.runLoggingT . handleOutput
```
jberthold, line 284: A bit puzzled that these function live in the `JsonRpc` space... I would have thought `Booster.Util`  is the most logical place for them :man_shrugging:

PR #514, opened 2024-02-20
Move Booster's `runServer` to `booster-dev`

geo2a:
The `Booster.JsonRpc.runServer` function has it's only call site in the `booster-dev` executable's `main`. I'm moving it there for consistency with `kore-rpc-booster`, which has similar code in `Proxy.hs`.

My primary motivation is to have the calls to `runHandleLoggingT` be in the executables' code, rather than leaking into Booster's library. It is going to be easier to set-up logging options with this layout.


goodlyrottenapple (approved):

jberthold (approved):
Good call! :+1: 
Reading through this code, I thought we could move the `Log.filterLogger levelFilter` idiom to where `runHandleLoggingT` lives (I would suggest `Booster.Util`)

geo2a:
Sounds good @jberthold, I'll make that change and merge.

geo2a:
I've decided to leave the logging filter alone since it captures a bunch of variables and factoring it out seems not worth the effort.

PR #515, opened 2024-02-20
Update nixpkgs+ghc

goodlyrottenapple:
Update nixpkgs to 23.11 and ghc to 9.6.4, inline with https://github.com/runtimeverification/haskell-backend/pull/3719


scripts/update-haskell-backend.sh
```diff
@@ -12,12 +12,4 @@ if [ $REV != $OLD_REV ]; then
...
   # Update the stack.yaml.lock file by running a stack command that accesses dependencies
   stack ls dependencies --test > /dev/null
-  # freeze cabal dependencies
-  $(dirname $0)/freeze-cabal-to-stack-resolver.sh
-  # since we use cabal.project as source of truth for nix builds, we need to provide
-  # a valid --sha256 hash for the haskell-backend library
-  # the easiest way to do this is to simply run nix, let it fail and substitute the
-  # correct sha back into cabal.project
-  SHA=$(nix build 2>&1 | grep -oP 'got:\s+\K.+') || true
-  sed -i -r "/haskell-backend.git/{n;s/tag:.*$/tag: $REV/;n;s|--sha256:.*$|--sha256: $SHA|}" cabal.project
```
goodlyrottenapple, line 22: We don't use cabal for dependency management at all any more, since we either use non-nix stack or nix, which actually relies on stack via stacklock2nix and provides a custom version of cabal with all the correct dependencies from stack.

dev-tools/kore-rpc-dev/Server.hs
```diff
@@ -271,6 +271,7 @@ mkKoreServer loggerEnv@Log.LoggerEnv{logAction} CLOptions{definitionFile, mainMo
                         { serializedModules = Map.singleton (ModuleName mainModuleName) sd
                         , receivedModules = mempty
                         , loadedDefinition
+                        , receivedModules = mempty
```
jberthold, line 274:
```suggestion
```
:bucket: :broom:

flake.nix
```diff
@@ -43,9 +44,11 @@
             };
           additionalHaskellPkgSetOverrides = hfinal: hprev:
             with final.haskell.lib; {
+              crypton-x509 = dontCheck hprev.crypton-x509;
               decision-diagrams = dontCheck hprev.decision-diagrams;
               fgl = dontCheck hprev.fgl;
-              haskeline = dontCheck hprev.haskeline;
+              # haskeline = dontCheck hprev.haskeline;
```
jberthold, line 50:
```suggestion
```

.github/workflows/test.yml
```diff
@@ -161,7 +156,8 @@ jobs:
         run: |
           nix --version
           export JQ=$(nix-build '<nixpkgs>' -A jq --no-link)/bin/jq
-          booster=$(nix build .#kore-rpc-booster --extra-experimental-features 'nix-command flakes' --print-build-logs --json | $JQ -r '.[].outputs | to_entries[].value')
+          nix build .#kore-rpc-booster --extra-experimental-features 'nix-command flakes' --print-build-logs
+          booster=$(nix build .#kore-rpc-booster --extra-experimental-features 'nix-command flakes' --json | $JQ -r '.[].outputs | to_entries[].value')
```
geo2a, line 160: I don't understand this change. Why is it better to separate the steps of building and extracting the executable path?

flake.nix
```diff
@@ -55,40 +57,35 @@
                     substituteInPlace library/Booster/VersionInfo.hs \
                       --replace '$(GitRev.gitHash)' '"${self.rev or "dirty"}"'
                   '';
-                  buildTarget = "kore-rpc-booster";
+                  # buildTarget = "kore-rpc-booster";
```
geo2a, line 60:
```suggestion
```

flake.nix
```diff
@@ -101,10 +98,20 @@
             booster-backend.pkgSet.hs-backend-booster;
         in {
           kore-rpc-booster = withZ3 pkgs hs-backend-booster "kore-rpc-booster";
+          kore-rpc-client = withZ3 pkgs hs-backend-booster "kore-rpc-client";
         });
 
       devShells = perSystem (system: {
-        # Separate cabal shell just for CI
+        # Separate fourmolu and cabal shells just for CI
```
geo2a, line 105:
```suggestion
        # Separate fourmolu and hlint shells just for CI
```

geo2a: Or scratch that, I see there's a cabal shell too...

geo2a (approved):

.github/workflows/test.yml
goodlyrottenapple: I believe @jberthold pointed out this can fail but the CI won't pick up on it. This way, it hopefully fails correctly.

geo2a (approved):

jberthold:
I believe we will have to also upgrade the `K` dependency in this PR, but IIRC to a custom version that adjusts the nix setup to avoid problems with our integration test :thinking:  ?

PR #524, opened 2024-02-26
Hotfix add labelled rule logging for kore

jberthold:
The rule logging used for `kore-rpc` in log level `RewriteKore` outputs rule locations but not rule labels for successful rewrites. This PR adds another log option which outputs each applied rule with its label.
This is important for rules added via `add-module` requests which don't usually have rule locations  (but do have labels).

### DIFF
--- a/tools/booster/Stats.hs
+++ b/tools/booster/Stats.hs
@@ -43,13 +43,13 @@ instance (Floating a, PrintfArg a, Ord a) => Pretty (RequestStats a) where
                 <+> withUnit stats.average
                 <+> parens ("+-" <+> withUnit stats.stddev)
                 <> ", range"
-                    <+> brackets (withUnit stats.minVal <> ", " <> withUnit stats.maxVal)
+                <+> brackets (withUnit stats.minVal <> ", " <> withUnit stats.maxVal)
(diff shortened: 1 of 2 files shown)


tools/booster/Stats.hs
```diff
@@ -43,13 +43,13 @@ instance (Floating a, PrintfArg a, Ord a) => Pretty (RequestStats a) where
...
                 <> ", range"
-                    <+> brackets (withUnit stats.minVal <> ", " <> withUnit stats.maxVal)
+                <+> brackets (withUnit stats.minVal <> ", " <> withUnit stats.maxVal)
             , "Total time in kore-rpc code:"
                 <+> withUnit stats.koreTotal
             , "Average time per request in kore-rpc code:"
                 <+> withUnit stats.koreAverage
                 <> ", max"
-                    <+> withUnit stats.koreMax
+                <+> withUnit stats.koreMax
```
jberthold, line 52: This was added by fourmolu-0.14.0.0, which I am supposed to use since the upgrade.

jberthold: ... and now CI's formatting run removed it :exploding_head: :man_shrugging:

geo2a (approved):

PR #531, opened 2024-02-29
Remove spurious newline from log messages

geo2a:
I did not realize  that `defaultLogStr` already adds a newline at the end --- no need to use `hPutStrLn`


jberthold (approved):
LGTM. Little details :-)
Maybe we should check/ catch such a thing in a test. A unit test might be possible now that the logging helper lives in the library. Otherwise a simple integration test that captures output and compares log strings.

PR #537, opened 2024-03-05
Remove code that handles outputting eventlog into a file

geo2a:
This functionality is not useful, and can be implemented on top of the JSON-logging infrastructure introduced in #516


goodlyrottenapple (approved):

jberthold (approved):
:fire:

PR #546, opened 2024-03-11
Add an OrTerm constructor

### LINKED ISSUE #545: Booster crashes if rule contains `\or`
Right now if you feed a K definition into the booster where a rule contains `\or` somewhere inside its LHS, the Haskell backend crashes and doesn't even try to rewrite anything. This is not ideal because the LLVM backend handles such cases and thus the Booster ought to be able to handle them in function rules in the case where the arguments to the function are fully concrete. It also ought not be particularly hard to handle the case of pattern matching on such patterns since it's simply a matter of reporting whether pattern matching succeeds on any of the disjuncts.

goodlyrottenapple:
Fixes #545. This allows internalising #Or terms, however we cannot match/unify on rules containing these for now.

### DIFF
--- a/library/Booster/Definition/Ceil.hs
+++ b/library/Booster/Definition/Ceil.hs
@@ -156,6 +156,7 @@ computeCeil term@(SymbolApplication symbol _ args)
                 argCeils <- concatMapM computeCeil args
                 pure $ (map Left $ splitBoolPredicates $ Predicate simplified) <> argCeils
 computeCeil (AndTerm l r) = concatMapM computeCeil [l, r]
+computeCeil (OrTerm l r) = concatMapM computeCeil [l, r] -- TODO: Is this too much of an overapproximation? I dont think it should be Ceil(l) \/ Ceil(r)...
 computeCeil (Injection _ _ t) = computeCeil t
 computeCeil (KMap _ keyVals rest) = do
     recArgs <- concatMapM computeCeil $ concat [[k, v] | (k, v) <- keyVals] <> maybeToList rest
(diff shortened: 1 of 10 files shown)


library/Booster/Definition/Ceil.hs
```diff
@@ -156,6 +156,7 @@ computeCeil term@(SymbolApplication symbol _ args)
                 argCeils <- concatMapM computeCeil args
                 pure $ (map Left $ splitBoolPredicates $ Predicate simplified) <> argCeils
 computeCeil (AndTerm l r) = concatMapM computeCeil [l, r]
+computeCeil (OrTerm l r) = concatMapM computeCeil [l, r] -- TODO: Is this too much of an overapproximation? I dont think it should be Ceil(l) \/ Ceil(r)...
```
dwightguth, line 159: if either L or R is non-bottom, then `Ceil(L) \/ Ceil(R)` should be `Top`. Similarly, if either L or R is non-bottom, then `L \/ R` should be non-bottom, which means `Ceil(L \/ R)` should be Top.

If both L and R are bottom, then `Ceil(L) \/ Ceil(R) == Ceil(L \/ R) == Bottom`

I'm actually more concerned about the line above: `Ceil(L) /\ Ceil(R)` can be `Top` even when `Ceil(L /\ R)` is `Bottom`.

goodlyrottenapple:
Closing this as it seems easier to split rules with an `#Or` internally and not trying to internalise terms with an or, as it doesn't seem make sense to send a configuration containing an `#Or` in an execute request anyway

PR #556, opened 2024-03-21
549 use get model booster implementation in proxy

### LINKED ISSUE #549: Use booster implementation for get-model endpoint
`get-model` is implemented in booster and should be used, rather than forwarding to the legacy code.

jberthold:
- Proxy uses booster endpoint for `get-model` requests first
- falls back to legacy-kore endpoint  on `Unknown` results or errors

The integration tests remain unchanged, they were already passing with `booster-dev` implementation before.

FIxes #549


jberthold:
No significant performance changes with `kontrol` (and unclear whether we have any `kontrol` tests for `get-model` anyway).

geo2a (approved):
I think we only call `get-model` in Kontrol when a proof fails, so this change does not affect performance.

PR #561, opened 2024-03-26
Return `rule-id`'s in execute response

### LINKED ISSUE #536: Branch `rule-id` and condition are missing from response
Related:
* runtimeverification/haskell-backend#3716
* runtimeverification/pyk#942

As of runtimeverification/haskell-backend#3716  `kore-rpc` returns an optional `rule-id` along with an optional `rule-substitution` and `rule-predicate`.  Ensure that the Booster returns the same data on branching.

Tests: runtimeverification/pyk#942.

goodlyrottenapple:
fixes #536 by returning the app;lied rule id  and `rule-substitution/predicate` for branching responses


test/rpc-integration/test-a-to-f/response-branching.booster-dev
```diff
@@ -0,0 +1,107 @@
+{
```
goodlyrottenapple, line 1: the rules now have a side-condition and the booster cannot currently handle such branching, so the booster-dev just aborts

jberthold: OK, so we should probably rename the test case to "aborting-on-side-condition", and have a different test input that triggers branching

jberthold (commented):
Not optimal that the `a-to-f` tests got more complicated. 
I wonder if a new test directory would have been better.

geo2a (approved):

jberthold (approved):

PR #566, opened 2024-04-02
More information on equations in logs

geo2a:
- for cached equation applications, specify which cache they came from
- if rule id is not available (usually for LLVM simplifications), emit the rule label instead.

### DIFF
--- a/library/Booster/Pattern/ApplyEquations.hs
+++ b/library/Booster/Pattern/ApplyEquations.hs
@@ -31,6 +31,7 @@ module Booster.Pattern.ApplyEquations (
     SimplifierCache,
 ) where
 
+import Control.Applicative (Alternative (..))
 import Control.Monad
 import Control.Monad.Extra (fromMaybeM, whenJust)
 import Control.Monad.IO.Class (MonadIO (..))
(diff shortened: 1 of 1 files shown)


library/Booster/Pattern/ApplyEquations.hs
```diff
@@ -259,6 +259,15 @@ isMatchFailure _ = False
...
 
+{- | Attempt to get an equation's unique id, falling back to it's label or eventually to UNKNOWN.
+  The fallbacks are useful in case of cached equation applications or the ones done via LLVM,
+  as neither of these categories have unique IDs.
+-}
+equationRuleIdWithFallbacks :: EquationMetadata -> Text
+equationRuleIdWithFallbacks metadata = case fmap getUniqueId metadata.ruleId of
+    Nothing -> fromMaybe "UNKNOWN" metadata.label
+    Just x -> x
+
```
goodlyrottenapple, line 270:
```suggestion
equationRuleIdWithFallbacks metadata =
    fromMaybe "UNKNOWN" $ fmap getUniqueId metadata.ruleId <|> metadata.label
```

goodlyrottenapple (approved):

PR #567, opened 2024-04-03
Add mx performance script

goodlyrottenapple:


scripts/performance-tests-mx.sh
```diff
@@ -0,0 +1,73 @@
...
+if [[ $MX_VERSION == "master" ]]; then
+  MX_VERSION=$(git name-rev --tags --name-only $(git rev-parse HEAD))
+else
+  MX_VERSION="${MX_VERSION//\//-}"
+fi
+
+git submodule update --init --recursive --depth 1 kmxwasm/k-src/mx-semantics/
+
+
+feature_shell "poetry -C kmxwasm install && make -C kmxwasm kbuild-haskell"
```
geo2a, line 58: Do we need to execute the same command in `master_shell`? 

Also, will the LLVM backend shared library be built by `make -C kmxwasm kbuild-haskell`? @virgil-serbanuta @bbyalcinkaya could you have a look at this PR and make sure it does what you'd expect?

goodlyrottenapple: yes, master shell is used fto compare our feature branch change to the current version of booster in mx

goodlyrottenapple: these steps are copied from the CI run in the mx repo:
https://github.com/runtimeverification/mx-backend/blob/872626e6a658bdbdc7914e5c1c01d19cffc3f11c/.github/workflows/test.yml#L96-L118
so hopefully should do the same thing....

jberthold: I would _guess_ that the `poetry install` and `make kbuild-haskell` are independent of the booster version, so we wouldn't  have to repeat that command (we don't at the moment). 
Better to have this confirmed by @virgil-serbanuta  or @bbyalcinkaya , though.

jberthold (approved):
LGTM. Will test this locally (Linux) but we can merge and iterate (not user-facing)

scripts/performance-tests-mx.sh
virgil-serbanuta: The kasmer tool builds the semantics at start (for the Haskell backend, for the LLVM backend, and, if `--booster` is passed to the tool, it also builds the llvm library). Tests do something similar. `make -C kmxwasm kbuild-haskell` should not be needed unless you need to run things outside our tool (or, perhaps, you just want to check that it compiles).

virgil-serbanuta: FWIW, we will probably add the option to use an already-built semantics soon, we can try to do that sooner if you need it.

jberthold: image

PR #573, opened 2024-04-10
Unify Match/Unify modules

### LINKED ISSUE #3785: Merge our `unification` and `matching` algorithms
At the moment we have essentially two copies of the "unification" algorithm in the booster:

https://github.com/runtimeverification/hs-backend-booster/blob/754e6ca645d96b0460355a4d61bdf42b3b4f6ff2/library/Booster/Pattern/Unify.hs#L172
https://github.com/runtimeverification/hs-backend-booster/blob/754e6ca645d96b0460355a4d61bdf42b3b4f6ff2/library/Booster/Pattern/Match.hs#L120

However, I would argue that our "unification" when applying rewrite rules is just a slightly different version of the matching algorithm we use for applying function and simplification rules. Namely, we expressly forbid any overlap of variables in the two terms and expressly check at the end of the unification algorithm that we in fact obtained a matching substitution, such that we have `subst(rule_LHS) = current_configuration`. Since there is no significant algorithmic difference between our "unification" and matching (match1 is a copy of unify1 with various tweaks), it would be better to merge the two as it will be much easier to see what exactly the differences between the two algorithms are. E.g. `match1` allows matching two terms `f1(...)` where `f1` isn't a constructor, whereas `unify1` does not allow this. Conversely, `unify1` works for a subset of map/set/list problems, whereas support for these is lacking in `match1`.

This issue has been raised before, but is more pertinent now that we are tackling the implication endpoint in the booster, which requires tweaks to either versions to e.g. lift the restriction that the variables in the two terms are completely disjoint. Instead of making yet another copy of essentially the same code, we should instead merge `unify1`/`match1` and put the differences between rule/function/implication matching in a single place, such that it is easy to see what the subtle differences (if any) are between these versions.

goodlyrottenapple:
Will re-open in the haskell backend, just keeping this for the commit history for now

fixes runtimeverification/haskell-backend#3785


goodlyrottenapple:
# Kontrol

| Test                                                                 | sam-unify-match-refactor time | master-754e6ca time | (sam-unify-match-refactor/master-754e6ca) time 
|----------------------------------------------------------------------|-------------------------------|---------------------|------------------------------------------------
| PlainPrankTest.test_stopPrank_notExistent()                          | 39.26                         | 53.04               | 0.7401960784313725                             
| PlainPrankTest.test_startPrank_zeroAddress_true()                    | 136.31                        | 178.45              | 0.763855421686747                              
| PlainPrankTest.test_prank_expectRevert()                             | 98.25                         | 121.9               | 0.805988515176374                              
| PlainPrankTest.test_prank_zeroAddress_true()                         | 159.16                        | 195.31              | 0.8149096308432747                             
| PlainPrankTest.test_startPrank_true()                                | 132.49                        | 160.96              | 0.8231237574552684                             
| PlainPrankTest.test_startPrankWithOrigin_true()                      | 134.12                        | 161.51              | 0.8304129775246115                             
| MockCallTestFoundry.testMockCallEmptyAccount()                       | 91.83                         | 102.98              | 0.8917265488444358                             
| FreshCheatcodes.test_bool()                                          | 45.15                         | 47.91               | 0.9423919849718222                             
| PlainPrankTest.testFail_startPrank_internalCall()                    | 48.46                         | 50.92               | 0.9516889238020424                             
| AddrTest.test_addr_true()-trace_options1                             | 32.67                         | 34.15               | 0.956661786237189                              
| PrankTestMsgSender.test_msgsender_setup()                            | 116.15                        | 120.96              | 0.9602347883597885                             
| AssertTest.checkFail_assert_false()                                  | 38.1                          | 39.65               | 0.960907944514502                              
| AccountParamsTest.testDealConcrete()-trace_options0                  | 32.08                         | 33.35               | 0.9619190404797601                             
| SetUpTest.testSetupData()                                            | 75.86                         | 78.8                | 0.9626903553299493                             
| BlockParamsTest.testChainId(uint256)                                 | 46.52                         | 48.24               | 0.9643449419568823                             
| AssumeTest.testFail_assume_false(uint256,uint256)                    | 67.06                         | 69.42               | 0.9660040334197638                             
| ERC721.sol                                                           | 48.41                         | 46.43               | 1.0426448416971785                             
| Storage.sol                                                          | 51.62                         | 49.2                | 1.0491869918699186                             
| src/tests/integration/test_foundry_prove.py::test_foundry_xml_report | 31.26                         | 29.77               | 1.0500503862949278                             
| AssertTest.test_assert_false()                                       | 85.75                         | 80.09               | 1.070670495692346                              
| MockCallTestFoundry.testMockNested()                                 | 432.57                        | 394.76              | 1.0957797142567636                             
| BMCLoopsTest.test_countdown_concrete()                               | 21.87                         | 19.3                | 1.133160621761658                              
| TOTAL                                                                | 1964.9499999999996            | 2117.1000000000004  | 0.9281328232015489

library/Booster/Pattern/Rewrite.hs
```diff
@@ -401,6 +403,8 @@ data RewriteFailed k
       UnificationIsNotMatch (RewriteRule k) Term Substitution
     | -- | A sort error was detected during unification
       RewriteSortError (RewriteRule k) Term SortError
+    | -- | A sort error was detected during unification
+      InternalMatchError Text
```
jberthold: 🍝 copy-pasta

test/llvm-integration/LLVM.hs
```diff
@@ -633,7 +633,7 @@ defSymbols =
                 , attributes =
                     SymbolAttributes
                         { collectionMetadata = Nothing
-                        , symbolType = TotalFunction
+                        , symbolType = Function Total
                         , isIdem = IsNotIdem
                         , isAssoc = IsNotAssoc
```
jberthold, line 638: Maybe better to just re-generate (using the repl, and then slightly edit and auto-format) this definition. It is not up to date any more with the source.

library/Booster/Pattern/UnifiedMatcher.hs
```diff
@@ -0,0 +1,831 @@
...
+data MatchResult
+    = -- | equal structure (constructors) after substitution (substitution goes both ways)
+      MatchSuccess Substitution
+    | -- | different constructors or domain values, or sort mismatch
+      MatchFailed FailReason
+    | -- | (other) cases that are unresolved (offending case in head position).
+      MatchIndeterminate (NonEmpty (Term, Term))
+    deriving stock (Eq, Show)
+
+data MatchType = Rule | Fun deriving (Eq)
```
jberthold: As discussed, maybe `Rewrite` and `Eval` rather than `Rule/Fun`. Or `Rule/Equn` ?

library/Booster/Pattern/UnifiedMatcher.hs
```diff
@@ -0,0 +1,831 @@
...
+            "Shared variables:" <+> hsep (map pretty $ Set.toList vs)
+        SubsortingError err ->
+            pretty $ show err
+        ArgLengthsDiffer t1 t2 ->
+            vsep ["Argument length differ", pretty t1, pretty t2]
+
+type Substitution = Map Variable Term
+
+{- | Attempts to find a simple unifying substitution for the given
+   terms. Only constructor symbols are considered (no functions).
```
jberthold:
```suggestion
   terms.
```

library/Booster/Pattern/UnifiedMatcher.hs
```diff
@@ -0,0 +1,831 @@
...
+
+   The returned substitution is oriented towards 'term1', i.e.,
+   prefers to replace its variables if given a choice.
+
+   This code calls `error` for internal errors (e.g., function
+   arguments that should not be possible), assuming the caller will
+   catch and handle those errors.
```
jberthold: The comments about orientation and error calls  are not correct any more, I guess.

library/Booster/Pattern/UnifiedMatcher.hs
```diff
@@ -0,0 +1,831 @@
...
+                        , mQueue = Seq.singleton (term1, term2) -- PriorityQueue.singleton (term1, term2) RegularTerm ()
+                        , mMapQueue = mempty
+                        , mIndeterminate = []
+                        , mSubsorts = Map.map snd sorts
+                        }
+
+data MatchState = State
+    { mSubstitution :: Substitution
+    , mTargetVars :: Set Variable
+    , mQueue :: Seq (Term, Term) -- PriorityQueue.HashPSQ (Term, Term) UnificationPriority () -- work queue with a priority
```
jberthold:
```suggestion
    , mQueue :: Seq (Term, Term) 
```

library/Booster/Pattern/UnifiedMatcher.hs
```diff
@@ -0,0 +1,831 @@
...
+        Seq.Empty -> case mapQueue of
+            Seq.Empty -> checkIndeterminate -- done
+            (term1, term2) :<| rest -> do
+                modify $ \s -> s{mMapQueue = rest}
+                match1 mType term1 term2
+                match mType
+        (term1, term2) :<| rest -> do
+            -- case PriorityQueue.minView queue of
+            --     Nothing -> checkIndeterminate -- done
+            --     Just ((term1, term2), _, _, rest) -> do
```
jberthold:
```suggestion
```

library/Booster/Pattern/UnifiedMatcher.hs
```diff
@@ -0,0 +1,831 @@
...
+    symbol1
+    sorts1
+    args1
+    symbol2
+    sorts2
+    args2
+        | symbol1.name /= symbol2.name =
+            failWith
+                ( DifferentSymbols (SymbolApplication symbol1 sorts1 args1) (SymbolApplication symbol2 sorts2 args2)
+                )
```
jberthold, line 378: This is a bit intricate... only correct with the invariant that we only call this case (with `Rule`) for `ConsApplication`s.
Maybe we could merge the `Rule` and `Fun` cases for `symbol1.name /= symbol2.name`  to make this more self-contained? (the `Rule` case shouldn't ever occur with function symbols, and would yield `Indeterminate` before)

library/Booster/Pattern/UnifiedMatcher.hs
```diff
@@ -0,0 +1,831 @@
...
+                                -- fully concrete list too short
+                                let surplusLeft = KList def [] $ Just (symb1, reverse tail1)
+                                failWith $ DifferentValues surplusLeft emptyList
+                            Just (Right tail2) -> do
+                                let newRight = KList def (reverse tail2) Nothing
+                                enqueueRegularProblem symb1 newRight
+        | -- mirrored case above: left list fully concrete, right one isn't
+          Nothing <- rest1
+        , Just _ <- rest2 =
+            matchLists def heads2 rest2 heads1 rest1 -- won't loop, will fail later if unification succeeds
```
jberthold, line 505: Not sure about this case... 
Since the intent is _matching_  this is probably bound to fail - the idea is that the pattern being fully concrete already means we cannot match a non-concrete list part in the subject but we might have to spell this out more explicitly.

library/Booster/Pattern/UnifiedMatcher.hs
```diff
@@ -0,0 +1,831 @@
...
+
+                    runMaybeT
+                        ( matchRemainderOfMapLists
+                            (toMapList (Map.toList patExtra) patRest)
+                            (toMapList (Map.toList subjExtra) subjRest)
+                        )
+                        >>= \case
+                            Just newProblems ->
+                                enqueueRegularProblems $ (Seq.fromList $ Map.elems commonMap) >< newProblems
+                            Nothing -> pure ()
```
jberthold: What about the `commonMap` in this case?

library/Booster/Pattern/UnifiedMatcher.hs
```diff
@@ -0,0 +1,831 @@
...
+
+data MapList
+    = ConstructorKey Term Term MapList
+    | Rest RestMapList
+
+pattern SingleConstructorKey :: Term -> Term -> MapList
+pattern SingleConstructorKey k v = ConstructorKey k v (Rest Empty)
+
+pattern SingleOtherKey :: Term -> Term -> MapList
+pattern SingleOtherKey k v = Rest (OtherKey k v Empty)
```
jberthold: I would prefer if we didn't bring in a whole new map matching algorithm in the same PR as the refactoring.
It should be possble to port the existing matching/"unification" code without too many adaptations, then we can still rewrite the map-matching algorithm.

library/Booster/Pattern/UnifiedMatcher.hs
```diff
@@ -0,0 +1,831 @@
...
+bindVariable mType var term = do
+    State{mSubstitution = currentSubst, mTargetVars = targets} <- get
+    case term of
+        -- Check if term is a variable, prefer target variables. Should
+        -- not happen given how we call it in the code above.
+        Var var2
+            | mType == Rule
+                && var2 `Set.member` targets
+                && not (var `Set.member` targets) ->
+                bindVariable mType var2 (Var var)
```
jberthold: An assumption of the prior "unification" code... we should probably remove `targets` completely from the code and not care whether or not `term` is a variable.

library/Booster/Pattern/UnifiedMatcher.hs
goodlyrottenapple: unfortunately I tried a few times to make a frankenstein monster out of the existing two versions but could not get the tests to pass in both instances. I also found one or two tests that are just wrong and others that had to be oriented correctly for matching

library/Booster/Pattern/UnifiedMatcher.hs
goodlyrottenapple: if we have a `Nothing` that means we added an indeterminate result, in which case I thought we don't want to continue anyway, unless we want to find more indeterminate cases in the `commonMap`, tho that just seems like busywork.

library/Booster/Pattern/UnifiedMatcher.hs
goodlyrottenapple: Yeah this was in the original unification code which appeared bidirectional, but we know this would eventually fail. Definitely should refactor, but maybe in a followup?

library/Booster/Pattern/UnifiedMatcher.hs
goodlyrottenapple: That being said, I have yet to run kevm/kontrol with this change so not too sure of it myself atm

goodlyrottenapple: Hmm perhaps i will just keep the two copies for now and try to introduce this in the followup

test/llvm-integration/LLVM.hs
goodlyrottenapple: should re-think this maybe parsing kore via TH? I would leave out of the scope of this PR at any rate...

library/Booster/Pattern/UnifiedMatcher.hs
jberthold: Would there be cases where the `commonMap` elements still cause an outright failure to match (as opposed to being indeterminate)?

library/Booster/Pattern/Rewrite.hs
```diff
@@ -397,10 +399,12 @@ data RewriteFailed k
       RuleConditionUnclear (RewriteRule k) Predicate
     | -- | A rewrite rule does not preserve definedness
       DefinednessUnclear (RewriteRule k) Pattern [NotPreservesDefinednessReason]
-    | -- | A unification produced a non-match substitution
+    | -- | A matching produced a non-match substitution
       UnificationIsNotMatch (RewriteRule k) Term Substitution
```
geo2a, line 403: Probably need to rename the `UnificationIsNotMatch` constructor to just `IsNotMatch`.

library/Booster/Pattern/UnifiedMatcher.hs
```diff
@@ -0,0 +1,839 @@
...
+match1 _       (AndTerm t1a t1b)                          t2@KList{}                                 = enqueueRegularProblem t1a t2 >> enqueueRegularProblem t1b t2
+match1 _       (AndTerm t1a t1b)                          t2@KSet{}                                  = enqueueRegularProblem t1a t2 >> enqueueRegularProblem t1b t2
+match1 _       (AndTerm t1a t1b)                          t2@ConsApplication{}                       = enqueueRegularProblem t1a t2 >> enqueueRegularProblem t1b t2
+match1 _       (AndTerm t1a t1b)                          t2@FunctionApplication{}                   = enqueueRegularProblem t1a t2 >> enqueueRegularProblem t1b t2
+match1 Rewrite (AndTerm t1a t1b)                          t2@Var{}                                   = enqueueRegularProblem t1a t2 >> enqueueRegularProblem t1b t2
+match1 Eval    t1@AndTerm{}                               t2@Var{}                                   = addIndeterminate t1 t2
+match1 Rewrite t1@DomainValue{}                           (AndTerm t2a t2b)                          = enqueueRegularProblem t1 t2a >> enqueueRegularProblem t1 t2b
```
geo2a, line 206: I like how this looks! I'd suggest a minor style change: let's add comments separating the groups of clauses for the different `Term` patterns: 

```suggestion
-- DomainValue
match1 Rewrite t1@DomainValue{} (AndTerm t2a t2b) = enqueueRegularProblem t1 t2a >> enqueueRegularProblem t1 t2b
```

library/Booster/Pattern/UnifiedMatcher.hs
```diff
@@ -0,0 +1,839 @@
...
+match1 _       t1@DomainValue{}                           t2@KMap{}                                  = failWith $ DifferentSymbols t1 t2
+match1 _       t1@DomainValue{}                           t2@KList{}                                 = failWith $ DifferentSymbols t1 t2
+match1 _       t1@DomainValue{}                           t2@KSet{}                                  = failWith $ DifferentSymbols t1 t2
+match1 _       t1@DomainValue{}                           t2@ConsApplication{}                       = failWith $ DifferentSymbols t1 t2
+match1 _       t1@DomainValue{}                           t2@FunctionApplication{}                   = addIndeterminate t1 t2
+match1 Rewrite t1@DomainValue{}                           (Var var2)                                 = matchVar Rewrite var2 t1
+match1 Eval    t1@DomainValue{}                           t2@Var{}                                   = addIndeterminate t1 t2
+match1 Rewrite t1@Injection{}                             (AndTerm t2a t2b)                          = enqueueRegularProblem t1 t2a >> enqueueRegularProblem t1 t2b
```
geo2a, line 217: As suggested above, consider adding spanners like this:

```suggestion
-- Injection
match1 Rewrite t1@Injection{}                             (AndTerm t2a t2b)                          = enqueueRegularProblem t1 t2a >> enqueueRegularProblem t1 t2b
```

library/Booster/Pattern/UnifiedMatcher.hs
```diff
@@ -0,0 +1,839 @@
...
+                then bindVariable matchType v (Injection source2 source1 trm2)
+                else failWith (DifferentSorts trm1 trm2)
+        | otherwise =
+            failWith (DifferentSorts (Injection source1 target1 trm1) (Injection source2 target2 trm2))
+{-# INLINE matchInj #-}
+
+----- Symbol Applications
+-- two constructors: fail if names differ, recurse
+
+----- Symbol Applications
```
geo2a, line 357:
```suggestion
----- Symbol Applications
```

library/Booster/Pattern/UnifiedMatcher.hs
```diff
@@ -0,0 +1,839 @@
...
+        emptyMap = KMap patDef [] Nothing
+
+        checkDuplicateKeys m@(KMap _ assocs _) =
+            let duplicates =
+                    Map.filter (> (1 :: Int)) $ foldr (flip (Map.insertWith (+)) 1 . fst) mempty assocs
+             in case Map.assocs duplicates of
+                    [] -> pure ()
+                    (k, _) : _ -> failWith $ DuplicateKeys k m
+        checkDuplicateKeys _ = pure ()
+matchMaps _ _ _ = undefined
```
geo2a, line 718: This is probably an overlooked case, or should it "never happen"? Either way, probably a good idea to return indeterminate here.

goodlyrottenapple: this is getting a refactor in the followup PR anyway

PR #575, opened 2025-10-01
Updating MacOs CI Build Runners

F-WRunTime:
MacOS 12 is outdated and deprecated by GH for public runners.  Internally we should target the latest OS


jberthold:
No neeed to update this repository, it is only here as an archive.

